PR #15, opened 2025-04-03
feat: use mask (*) instead of parameter names while matching rest operations during comparison

CountRedClaw:
Closes [7](https://github.com/Netcracker/qubership-apihub/issues/7)


src/components/compare/compare.operations.ts
```diff
@@ -343,21 +363,20 @@ async function compareCurrentApiType(
...
-type HandleType = typeof HANDLE_TYPE_FULLY_ADDED | typeof HANDLE_TYPE_FULLY_REMOVED
-
 const createPairOperationsMap = (currGroupSlug: string, prevGroupSlug: string, currentOperations: ResolvedOperation[], previousOperations: ResolvedOperation[]): Record<string, { previous?: ResolvedOperation; current: ResolvedOperation }> => {
 
   const operationsMap: Record<string, { previous?: ResolvedOperation; current: ResolvedOperation }> = {}
 
   for (const currentOperation of currentOperations) {
-    operationsMap[takeSubstringIf(!!currGroupSlug, currentOperation.operationId, currGroupSlug.length)] = { current: currentOperation }
+    // todo
+    const normalizedOperationId = currentOperation.apiType === REST_API_TYPE ? slugify(`${currentOperation.metadata.path}-${currentOperation.metadata.method}`, [], IGNORE_PATH_PARAM_UNIFIED_PLACEHOLDER) : currentOperation.operationId
```
JayLim2, line 372: maybe extract to utility "normalizeOperationId"?

src/components/compare/compare.utils.ts
```diff
@@ -86,28 +94,40 @@ export function getOperationTypesFromTwoVersions(
...
-  const resolvedHashMap = { ...operationTypes.find(({ apiType: type }) => type === currentApiType)?.operations || {} }
+): [ResolvedVersionOperationsHashMap, OperationIdentityMap] {
   const { buildType, currentGroup, previousGroup } = ctx.config
+  const currentApiTypeOperations = operations.filter(({ apiType }) => apiType === currentApiType)
+
+  const resolvedHashMap: ResolvedVersionOperationsHashMap = {}
+  const normalizedToOriginalOperationIdMap: Record<NormalizedOperationId | OperationId, OperationId> = {}
+
+  for (const { operationId, apiType, metadata, dataHash } of currentApiTypeOperations) {
+    const normalizedOperationId = apiType === REST_API_TYPE ? slugify(`${metadata.path}-${metadata.method}`, [], IGNORE_PATH_PARAM_UNIFIED_PLACEHOLDER) : operationId
```
JayLim2, line 112: and here

src/components/compare/compare.operations.ts
```diff
@@ -343,21 +363,20 @@ async function compareCurrentApiType(
...
+    operationsMap[takeSubstringIf(!!currGroupSlug, normalizedOperationId, currGroupSlug.length)] = { current: currentOperation }
   }
 
   for (const previousOperation of previousOperations) {
-    const prevOperationId = takeSubstringIf(!!prevGroupSlug, previousOperation.operationId, prevGroupSlug.length)
+    // todo
+    const normalizedOperationId = previousOperation.apiType === REST_API_TYPE ? slugify(`${previousOperation.metadata.path}-${previousOperation.metadata.method}`, [], IGNORE_PATH_PARAM_UNIFIED_PLACEHOLDER) : previousOperation.operationId
```
JayLim2, line 378: and here

JayLim2 (changes requested):

src/components/compare/compare.operations.ts
CountRedClaw: extracted to createNormalizedOperationId

src/components/compare/compare.utils.ts
CountRedClaw: extracted to createNormalizedOperationId

src/components/compare/compare.operations.ts
CountRedClaw: extracted to createNormalizedOperationId

CountRedClaw:
@iugaidiana,
In order to get operations (/api/v2/packages/{packageId}/versions/{version}/{apiType}/operations), api-processor logic needs to know the api types that exist in version first.
But to get these api types, we need to pass either includeOperations or includeSummary flag as a query parameter in /api/v2/packages/{packageId}/versions/{version}. 
But the only field api-processor logic needs is an apiType field (in operationTypes array), so both of the flags are not suitable.

src/components/compare/compare.utils.ts
```diff
@@ -87,27 +97,36 @@ type OperationIdWithoutGroupPrefix = string
...
-): ResolvedVersionOperationsHashMap {
-  const resolvedHashMap = { ...operationTypes.find(({ apiType: type }) => type === currentApiType)?.operations || {} }
+): [ResolvedVersionOperationsHashMap, OperationIdentityMap] {
   const { buildType, currentGroup, previousGroup } = ctx.config
+  const resolvedHashMap: ResolvedVersionOperationsHashMap = {}
+  const normalizedToOriginalOperationIdMap: Record<NormalizedOperationId | OperationId, OperationId> = {}
+
+  for (const operation of operations) {
+    const { operationId, dataHash } = operation
+    const normalizedOperationId = apiBuilder.createNormalizedOperationId?.(operation) ?? operationId
```
JayLim2, line 111: I saw such function in rest.operations.ts. And it was standalone function. 
Why do we have different function with the same name in ApiBuilder and why do we provide ApiBuilder into utility function?

JayLim2 (changes requested):

src/components/compare/compare.operations.ts
```diff
@@ -343,21 +355,27 @@ async function compareCurrentApiType(
...
 
   for (const currentOperation of currentOperations) {
-    operationsMap[takeSubstringIf(!!currGroupSlug, currentOperation.operationId, currGroupSlug.length)] = { current: currentOperation }
+    const normalizedOperationId = apiBuilder.createNormalizedOperationId?.(currentOperation) ?? currentOperation.operationId
+    operationsMap[takeSubstringIf(!!currGroupSlug, normalizedOperationId, currGroupSlug.length)] = { current: currentOperation }
   }
 
   for (const previousOperation of previousOperations) {
-    const prevOperationId = takeSubstringIf(!!prevGroupSlug, previousOperation.operationId, prevGroupSlug.length)
+    const normalizedOperationId = apiBuilder.createNormalizedOperationId?.(previousOperation) ?? previousOperation.operationId
```
JayLim2, line 377: I saw such function in rest.operations.ts. And it was standalone function.
Why do we have different function with the same name in ApiBuilder and why do we provide ApiBuilder into utility function?

JayLim2 (changes requested):

src/components/compare/compare.utils.ts
CountRedClaw: compare.utils.ts contains common logic for all api types, and I didn't want to check the api type to decide whether there's a need to create normalized operationId in all 3 places. 
In my opinion, it is more convenient approach, especially since we already have the ApiBuilder abstraction

JayLim2: code is not duplicated - function in builder is provided with reference, not copy-paste

JayLim2 (approved):

PR #18, opened 2025-04-25
fix: do not fail publish on description override in OAS 3.0.x

b41ex:
* split 3 types of possible reference errors
* only fail publication for REF_NOT_FOUND or REF_NOT_VALID_FORMAT
* add corresponding test


b41ex:
Merged separately

PR #35, opened 2025-08-25
feat: support for reusable pathItems definitions in components for OAS 3.1

makeev-pavel:


src/apitypes/rest/rest.operation.ts
```diff
@@ -257,19 +258,60 @@ const createSingleOperationSpec = (
...
+    components: {
+      ...takeIfDefined({ securitySchemes }),
+    },
+  }
+
+  if (pathData && '$ref' in pathData && pathData.$ref) {
+    return {
+      ...specBase,
+      paths: {
+        [path]: { ...pathData },
```
CountRedClaw, line 298: why spread?

CountRedClaw (commented):
Please add a test in 'Document Group test' that makes sure that we are properly getting rid of operations that are not included in the group. Also it would be nice to check that we can handle mixed use of the regular pathItems and the referenced ones in one document.

src/apitypes/rest/rest.operation.ts
makeev-pavel: fixed

makeev-pavel:
> Please add a test in 'Document Group test' that makes sure that we are properly getting rid of operations that are not included in the group. Also it would be nice to check that we can handle mixed use of the regular pathItems and the referenced ones in one document.

added 2 tests

test/projects/merge-operations/case2/1.yaml
b41ex, line 1: Please, come up with a meaningful names for the cases, not just case1, case2, case3.

src/apitypes/rest/rest.operation.ts
```diff
@@ -257,19 +258,60 @@ const createSingleOperationSpec = (
 ): TYPE.RestOperationData => {
   const pathData = document.paths[path] as OpenAPIV3.PathItemObject
 
-  return {
+  const resolveRefPathItem = (ref: string): OpenAPIV3.PathItemObject | null => {
```
b41ex, line 261: We are essentially replicating reference resolution logic from api-unifier, which seems like a bad idea.
I don't understand how complex cases are support (it seems they are not), e.g.:
- reference chains
- reference cycles

If we replicate reference resolution logic for path items in api-processor, we need to support and add tests for all cases we are checking here:
https://github.com/Netcracker/qubership-apihub-api-unifier/blob/83f9c5c75a062b6edbd4cea8ed1494312aa64dfc/test/oas/resolve-ref.reference-object.test.ts#L26

Or (better) to come up with a solution which reuses reference resolution from api-unifier.

b41ex (changes requested):

src/strategies/document-group.strategy.ts
```diff
@@ -116,48 +110,106 @@ function extractDocumentData(versionDocument: VersionDocument): OpenAPIV3.Docume
...
+      const updatedPathItem = buildPath(
+        sourceDocument,
+        path,
+        inferredMethod,
+        commonPathProps,
+        pathItemRef,
+      )
+
+      resultDocument.paths[path] = {
+        ...(resultDocument.paths[path] || {}),
```
CountRedClaw, line 160: what is this line for? seems like removing it does not break any tests

CountRedClaw: looks like a test for multiple methods in one path case is missing

test/projects/merge-operations/case2/1.yaml
makeev-pavel: Changed the names. Pls check

src/apitypes/rest/rest.operation.ts
makeev-pavel: Yes, this approach was a dead end. Fixed it on the api-unifier.

src/apitypes/rest/rest.operation.ts
```diff
@@ -226,8 +242,70 @@ export const calculateSpecRefs = (sourceDocument: unknown, normalizedSpec: unkno
...
+
     setValueByPath(resultSpec, matchResult.path, component)
   })
+
+  if (operations?.length) {
+    resolveComponentsPathItemOperationSpec(resultSpec, normalizedSpec, operations)
+  }
+}
+
+export function resolveComponentsPathItemOperationSpec(
```
b41ex, line 254: Need a better name. It seems to me that this method reduces path items in components to a given set of operations.
Also sourceDocument argument name is misleading, since it is a result spec to which we post updated path items.

src/apitypes/rest/rest.operation.ts
```diff
@@ -257,15 +340,20 @@ const createSingleOperationSpec = (
 ): TYPE.RestOperationData => {
   const pathData = document.paths[path] as OpenAPIV3.PathItemObject
 
+  const isContainsRef = !!pathData.$ref
```
b41ex, line 343: isContainsRef -> isRefPathData ?

src/apitypes/rest/rest.operation.ts
```diff
@@ -257,15 +340,20 @@ const createSingleOperationSpec = (
...
-      [path]: {
-        ...extractCommonPathItemProperties(pathData),
-        [method]: { ...pathData[method] },
-      },
+      [path]: isContainsRef
+        ? pathData
+        : {
+          ...extractCommonPathItemProperties(pathData),
+          [method]: { ...pathData[method] },
+          ...(refFlag ? { [INLINE_REFS_FLAG]: refFlag } : {}),
```
b41ex, line 355: I would propose some utility method like `copySymbolIfDefined`, instead of separately calculating refFlag and separately copying symbol.

b41ex (changes requested):

test/hash.test.ts
```diff
@@ -0,0 +1,40 @@
...
+      )
+      expect(operationFile).not.toBeNull()
+
+      const operationData = JSON.parse(operationFile!)
+
+      expect(operationData.operations.length).toBe(2)
+      expect(operationData.operations[0]).not.toEqual(operationData.operations[1])
+    }
+
+    test('should be different hashes for each operation', async () => {
```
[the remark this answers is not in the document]
makeev-pavel: all expects in publishAndGetHash

src/apitypes/rest/rest.operation.ts
makeev-pavel: I have given a new name "reduceComponentPathItemsToOperations"

src/apitypes/rest/rest.operation.ts
makeev-pavel: as you wish

src/strategies/document-group.strategy.ts
```diff
@@ -116,48 +109,81 @@ function extractDocumentData(versionDocument: VersionDocument): OpenAPIV3.Docume
...
-      const operationPath = basePath + path
+      const methodData = normalizedPathItem[httpMethod]
+      const basePath = getOperationBasePath(
+        methodData?.servers ||
+        sourcePathItem?.servers ||
+        sourceDocument?.servers ||
+        [],
+      )
 
+      const operationPath = basePath + path
```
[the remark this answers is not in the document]
makeev-pavel: done

src/utils/builder.ts
```diff
@@ -29,6 +29,8 @@ import {
 import { API_KIND } from '../consts'
 import { Diff, DiffType } from '@netcracker/qubership-apihub-api-diff'
 import { JsonPath } from '@netcracker/qubership-apihub-json-crawl'
+import { parseRef } from '@netcracker/qubership-apihub-api-unifier'
```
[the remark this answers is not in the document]
makeev-pavel: yes, deleted

src/utils/objects.ts
```diff
@@ -65,3 +65,20 @@ export const isSymbol = (value: unknown): value is symbol => {
 export const isObject = (value: unknown): value is Record<string | symbol, unknown> => {
   return typeof value === 'object' && value !== null
 }
+
+export const getSymbolValueIfDefined = <T extends object>(
```
b41ex, line 69: I do not get, what is the benefit to have this method compared to just using `obj[symbol]`?

src/utils/objects.ts
```diff
@@ -65,3 +65,20 @@ export const isSymbol = (value: unknown): value is symbol => {
...
+  obj: T,
+  symbol: symbol,
+): unknown => {
+  const symbolObj = obj as { [key: symbol]: unknown }
+
+  return symbol in symbolObj ? symbolObj[symbol] : undefined
+}
+
+export const copySymbolIfDefined = <T extends object>(
```
b41ex, line 78: With this signature it does not really copy the symbol? If we want to have logic as is- need different name. `pickSymbolIfDefined`? Or may be something better.

src/apitypes/rest/rest.operation.ts
```diff
@@ -212,10 +227,13 @@ export const calculateSpecRefs = (sourceDocument: unknown, normalizedSpec: unkno
       return
     }
     const componentName = matchResult.grepValues[grepKey].toString()
-    const component = getKeyValue(sourceDocument, ...matchResult.path)
+    let component = getKeyValue(sourceDocument, ...matchResult.path) as Record<string, unknown>
     if (!component) {
       return
     }
+    if (isObject(component)) {
```
b41ex, line 234: Why we didn't need this before, but need now?
Commenting this block does not fail any tests.

src/apitypes/rest/rest.operation.ts
```diff
@@ -226,8 +244,70 @@ export const calculateSpecRefs = (sourceDocument: unknown, normalizedSpec: unkno
...
+export function reduceComponentPathItemsToOperations(
+  resultSpec: RestOperationData,
+  normalizedDocument: RestOperationData,
+  operations: OperationId[],
+): void {
+  const { paths } = normalizedDocument
+
+  for (const path of Object.keys(paths)) {
+    const sourcePathItem = paths[path] as OpenAPIV3.PathItemObject
+    if (!isNonNullObject(sourcePathItem)) {
```
b41ex, line 265: Why this check? No tests are failing if we remove it. In my understanding, we can't have non-object or `null` path in normalized document, api-unifier won't allow it, so, this is just excessive check that complicates code understanding.

src/apitypes/rest/rest.operation.ts
```diff
@@ -226,8 +244,70 @@ export const calculateSpecRefs = (sourceDocument: unknown, normalizedSpec: unkno
...
+      continue
+    }
+    const refs = getSymbolValueIfDefined(sourcePathItem, INLINE_REFS_FLAG) as string[] | undefined
```
b41ex, line 268: If would suggest extract line 268 to 272 to some utility method with meaningful name, like `getPathItemComponent`, this would simplify reading

src/apitypes/rest/rest.operation.ts
```diff
@@ -226,8 +244,70 @@ export const calculateSpecRefs = (sourceDocument: unknown, normalizedSpec: unkno
...
+    if (!refs || refs?.length === 0) {
+      continue
+    }
+    const { jsonPath } = parseRef(refs[0])
+    if (!jsonPath) {
+      continue
+    }
+
+    const valueByPath = getValueByPath(resultSpec, jsonPath) as OpenAPIV3.PathItemObject
```
b41ex, line 277: `pathItemComponent` would be more meaningful name here

src/apitypes/rest/rest.operation.ts
```diff
@@ -226,8 +244,70 @@ export const calculateSpecRefs = (sourceDocument: unknown, normalizedSpec: unkno
...
+    const sourcePathItem = paths[path] as OpenAPIV3.PathItemObject
+    if (!isNonNullObject(sourcePathItem)) {
+      continue
+    }
+    const refs = getSymbolValueIfDefined(sourcePathItem, INLINE_REFS_FLAG) as string[] | undefined
+    if (!refs || refs?.length === 0) {
+      continue
+    }
+    const { jsonPath } = parseRef(refs[0])
+    if (!jsonPath) {
```
b41ex, line 273: `pathItemComponentJsonPath` would be a better name

src/apitypes/rest/rest.operation.ts
```diff
@@ -281,3 +366,16 @@ export const extractCommonPathItemProperties = (
   ...takeIfDefined({ servers: pathData?.servers }),
   ...takeIfDefined({ parameters: pathData?.parameters }),
 })
+
+function isValidHttpMethod(method: string): method is OpenAPIV3.HttpMethods {
+  return (Object.values(OpenAPIV3.HttpMethods) as string[]).includes(method)
```
b41ex, line 371: Maybe let's use pre-computed set to avoid potential performance problems? Like 

```TypeScript
const HTTP_METHODS_SET = new Set(Object.values(OpenAPIV3.HttpMethods) as string[]);

function isValidHttpMethod(method: string): method is OpenAPIV3.HttpMethods {
  return HTTP_METHODS_SET.has(method);

```

src/apitypes/rest/rest.operation.ts
```diff
@@ -281,3 +366,16 @@ export const extractCommonPathItemProperties = (
...
+}
+
+export function calculateOperationId(
```
b41ex, line 374: There's existing method `createNormalizedOperationId`- what is the difference? If we need 2 different methods- let's at least put them next to each other.

b41ex (changes requested):

src/utils/objects.ts
makeev-pavel: The logic is simple - type conversion and validation in a separate place.

src/utils/objects.ts
makeev-pavel: here we get a symbol with a value, I would just call it getSymbolWIthValueIfDefined.

src/apitypes/rest/rest.operation.ts
makeev-pavel: I don't think it's necessary anymore, we're copying it somewhere else now.

src/apitypes/rest/rest.operation.ts
makeev-pavel: i created getPathItemComponentJsonPath 

because in lines 268 to 272 we get exactly the path, not the PathItem itself.

We will continue to use this path.

src/apitypes/rest/rest.operation.ts

src/apitypes/rest/rest.operation.ts
makeev-pavel: calculateOperationId and createNormalizedOperationId have difference result for operations. 
Example: 
calculateOperationId get "v1-pets-petid-get"
createNormalizedOperationId get "v1-pets-*-get"

Moved to utils

src/utils/operations.utils.ts
```diff
@@ -82,3 +83,21 @@ export function isOperationRemove(operationDiff: Diff): boolean {
   // length is 2 because json path has view like ['paths', '/test/element']
   return operationDiff.action === DiffAction.remove && !!matchPaths(operationDiff.beforeDeclarationPaths, [[OPEN_API_PROPERTY_PATHS, PREDICATE_ANY_VALUE]])
 }
+
+export const isValidHttpMethod = (method: string): method is OpenAPIV3.HttpMethods => {
+  return HTTP_METHODS_SET.has(method)
+}
+
+export const createNormalizedOperationId: OperationIdNormalizer = (operation) => {
```
b41ex, line 91: Let's also align function names to avoid possible confusion. `calculate` is better for this case, so, let's rename existing method to `calculateNormalizedOperationId`

src/utils/objects.ts
b41ex: Another proposal- `extractSymbolProperty`?

b41ex (changes requested):
We have a logic that reduces path items to particular operation in a single-operation specification during `build` build type, but there's no corresponding test. 
I believe we should add one.
There's a similar test for `reducedSourceSpecifications` build type ('should have documents stripped of operations other than from provided group'), but not `build` build type.

PR #41, opened 2025-09-17
chore: fix super-linter reported issues

makeev-pavel:


test/projects/bugs/ref-in-invalid-spec.yaml
b41ex, line 1: Based on a file name, may be this spec should remain non-valid?

b41ex (changes requested):

PR #56, opened 2025-12-01
fix: date parsing from yaml

### LINKED ISSUE #402: [FE] APIHUB date time field Parsing error
### Describe the bug

APIHUB fails to correctly parse date-time string values in OpenAPI YAML files when the values are not enclosed in quotes. The parser incorrectly interprets unquoted date-time strings as empty objects {} instead of strings, causing validation errors.
[bug_ticket_example.zip](https://github.com/user-attachments/files/23019179/bug_ticket_example.zip)

### To Reproduce

1. Publish versions (attached above)
2. Observe the validation errors in the Operation tab
3. Check that testConnectionDate is parsed as {} instead of "2022-03-10T16:15:50Z"

**ER: Unquoted date-time strings like 2022-03-10T16:15:50Z should be parsed as strings without validation errors, maintaining the actual date-time value in the testConnectionDate field.
AR: Unquoted date-time strings are parsed as empty objects {}, causing validation errors "testConnectionDate" property type must be string and losing the actual date-time value.**

iurii-golovinskii:


JayLim2 (changes requested):
See comment in https://github.com/Netcracker/qubership-apihub-api-diff/pull/53

JayLim2 (approved):

makeev-pavel (approved):

vOrigins (approved):

PR #59, opened 2025-12-05
feat: add a hash cache to improve performance of build result serialization

### LINKED ISSUE #441: Changelog publishing for large GraphQL specification does not complete
When publishing large GraphQL specification from UI the process is not complete in a reasonable time. At the same time according to log messages all activities in building the changelog is complete.
The same specification could be published successfully via build-task-consumer.

makeev-pavel:


src/apitypes/rest/rest.operation.ts
```diff
@@ -79,7 +80,8 @@ export const buildRestOperation = (
   basePath: string,
   notifications: NotificationMessage[],
   config: BuildConfig,
-  componentsHashMap: Map<string, string>,
+  objectHashCache: ObjectHashCache,
+  componentsHashMap: ObjectHashCache,
```
JayLim2: maybe `componentsHashCache` if it is the same type? and then what means `objectHashCache` - what's kind of objects?

src/utils/builder.ts
```diff
@@ -89,7 +88,10 @@ export const jsonPathToString = (path: ObjPath): string => {
     .join('/')
 }
 
-export const removeObjectDuplicates = (originalArray: any[], by: string | ((item: any) => any)): any[] => {
+export const removeObjectDuplicates = (
+  originalArray: any[],
+  by: string | ((item: any) => any),
+): any[] => {
```
JayLim2: get rid of `any`

src/utils/hashes.ts
```diff
@@ -17,23 +17,29 @@
 import objectHash, { NotUndefined } from 'object-hash'
 import { isObject } from './objects'
 
-const objectHashCache = new WeakMap<object, string>()
+export type ObjectHashCache = WeakMap<object, string>
 
-export function calculateObjectHash(value: NotUndefined): string {
+export const calculateObjectHash = (value: NotUndefined): string => {
```
JayLim2: function declaration is better because JS loads declaration first, and executes everything only after that

src/utils/hashes.ts
```diff
@@ -17,23 +17,29 @@
...
+  value: unknown,
+  objectHashCache?: ObjectHashCache,
+): string => {
+  if (value === undefined) return ''
 
-    const hash = objectHash(value, { algorithm: 'md5' })
-    objectHashCache.set(value as object, hash)
-    return hash
+  if (!objectHashCache || !isObject(value)) {
+    return calculateObjectHash(value)
```
JayLim2: I read that like `if "value" is not an object, calculate object hash`. Is it correct title of function? Or you mean that hash is an object? In that case, I would suggest to rename it to `objectiveHash` or `complexHash` or smth like that.

src/utils/operations.utils.ts
```diff
@@ -119,3 +121,14 @@ export const createSerializedInternalDocument = (document: VersionDocument, effe
...
   versionInternalDocument.serializedVersionDocument = serializeDocument(denormalize(effectiveDocument, options) as ApiDocument)
 }
+
+export const calculateOperationHash = (
+  isOperation: boolean,
+  value: NotUndefined,
+  objectHashCache: ObjectHashCache,
+): string | undefined => {
+  if (isOperation) {
+    return undefined
```
JayLim2: the same: `function calculateOperationHash returns "undefined" is this is an operation, and calculates hash otherwise`. Weird, in my opinion

src/utils/transformToDto.ts
```diff
@@ -54,7 +54,7 @@ export function toChangeMessage(diff: ArrayType<Diff[]>, logError: (message: str
       const {
         afterDeclarationPaths,
       } = newDiff
-      const afterValueNormalized = (newDiff as Record<symbol, unknown>)[AFTER_VALUE_NORMALIZED_PROPERTY]
+      const afterValueNormalized  = (newDiff as Record<symbol, unknown>)[AFTER_VALUE_NORMALIZED_PROPERTY]
```
JayLim2: redundant whitespace

JayLim2 (changes requested):

src/apitypes/graphql/graphql.operation.ts
```diff
@@ -72,14 +73,13 @@ export const buildGraphQLOperation = (
...
 
-      const hash = isOperation ? undefined : calculateObjectHash(value)
       result.push({
         declarationJsonPaths,
         ...takeIfDefined({ description }),
         ...takeIfDefined({ deprecatedInfo: deprecatedReason }),
         ...takeIf({ [isOperationDeprecated]: true }, isOperation),
         deprecatedInPreviousVersions: config.status === VERSION_STATUS.RELEASE ? [version] : [],
-        ...takeIfDefined({ hash: hash }),
+        ...takeIfDefined({ hash: calculateOperationHash(isOperation, value, objectHashCache) }),
```
b41ex: I think it is better to keep special handling for operations here, as it was before, rather then incapsulating it into function, because function `calculateOperationHash` with argument `isOperation` looks strange. So, imo no need to specialized `calculateOperationHash` function, it could be just generic hash calculation finction.

src/utils/hashes.ts
```diff
@@ -17,23 +17,29 @@
...
+}
 
-  // cache by reference for objects and functions
-  //TODO: think of proper way/place to implement caching
-  if (isObject(value)) {
-    const cached = objectHashCache.get(value as object)
-    if (cached !== undefined) {
-      return cached
-    }
+export const calculateHash = (
```
b41ex: Why we are introducing second function? `calculateObjectHash` and `calculateHash`. If we have two function- there's always a question- when to use one or another, and I don't see an easy answer here.

src/apitypes/rest/rest.operation.ts
```diff
@@ -234,14 +235,7 @@ export const calculateSpecRefs = (
...
     if (models && !models[componentName] && isComponentsSchemaRef(matchResult.path)) {
-      let componentHash = componentsHashMap?.get(componentName)
-      if (componentHash) {
-        models[componentName] = componentHash
-      } else {
-        componentHash = calculateObjectHash(component)
-        componentsHashMap?.set(componentName, componentHash)
-        models[componentName] = componentHash
-      }
+      models[componentName] = calculateHash(component, componentsHashMap)
```
b41ex: We are introducing change here- previously we have cached by `componentName`, now we are caching by the object itself- are you sure it has the same behavior in all cases?

Upd: actually I think this change indeed breaks the intended behavior- during `build` we are currently processing per-operation specifications, and in these specifications instances of models will be different (so new cache will not work), but the names will be the same (so, old caching approach worked).

src/utils/calculateDiffId.ts
```diff
@@ -38,8 +38,8 @@ export function calculateDiffId(diff: Diff): string {
     action,
     previousDeclarationJsonPaths,
     currentDeclarationJsonPaths,
-    previousValueHash: beforeValueNormalized !== undefined ? calculateObjectHash(beforeValueNormalized) : '',
-    currentValueHash: afterValueNormalized !== undefined ? calculateObjectHash(afterValueNormalized) : '',
+    previousValueHash: calculateHash(beforeValueNormalized),
```
b41ex, line 43: Why no hash cache here? This place is a known performance bottleneck, we should definitely use cache here.

src/apitypes/rest/rest.operation.ts
```diff
@@ -79,7 +80,8 @@ export const buildRestOperation = (
   basePath: string,
   notifications: NotificationMessage[],
   config: BuildConfig,
-  componentsHashMap: Map<string, string>,
+  objectHashCache: ObjectHashCache,
```
b41ex: I would propose to call this field `normalizedObjectsHashCache` to to highlight what it can be used for.

b41ex (changes requested):

src/utils/calculateDiffId.ts
makeev-pavel: This is necessary for removeObjectDuplicates. We decided that there can be no identical instances here.

src/utils/hashes.ts
makeev-pavel: This is necessary for jest.spyOn to work

src/apitypes/rest/rest.operation.ts
makeev-pavel: renamed to normalizedObjectsHashCache

src/utils/builder.ts
makeev-pavel: rolled back the changes

src/utils/hashes.ts
makeev-pavel: changed the name of the function. Now its calculateMd5Hash. Since we use objectHash with the 'md5' algorithm and can get the hash not only from the object.

src/utils/operations.utils.ts
makeev-pavel: deleted

src/utils/transformToDto.ts
makeev-pavel: deleted

src/apitypes/graphql/graphql.operation.ts
makeev-pavel: I returned it as it was + cache

src/utils/hashes.ts
makeev-pavel: I changed the name to make it clear. One function is useful, it is needed both for tests and to avoid specifying several times which algorithm to calculate the hash with

src/apitypes/rest/rest.operation.ts
makeev-pavel: Yes, I rolled back the changes

src/apitypes/rest/rest.operation.ts

src/utils/hashes.ts
b41ex: It could still be confusing. I see that it is used 2 times inside hashes.ts and in tests. 
Can we replace spying on this function to spying on `object hash` library original function? If we do- I would say that it is better to just get rid of `calculateMd5Hash` to avoid confusion which to use. 
If we can't and we need to export this (essentially) internal function to use in tests- I would propose to add underscore in the beginning to convey that it is internal function, so: `_calculateMd5Hash`

src/utils/calculateDiffId.ts
b41ex: Please add a comment here that calculate diffId is used when deduplicating diffs from different document, hence hash cache by instance won't work here, hence we just calculate hash without using cache.

src/builder.ts
```diff
@@ -222,6 +225,7 @@ export class PackageVersionBuilder implements IPackageVersionBuilder {
       versionDeprecatedResolver: this.versionDeprecatedResolver.bind(this),
       versionDocumentsResolver: this.versionDocumentsResolver.bind(this),
       rawDocumentResolver: this.rawDocumentResolver.bind(this),
+      objectHashCache: this.normalizedObjectsHashCache,
```
b41ex: It makes sense to change the name in all places `objectHashCache` -> `normalizedObjectsHashCache`

b41ex (changes requested):

src/apitypes/rest/rest.operation.ts
JayLim2: we discussed it yesterday, it doesn't have sense what normalized objects are. everything in JS is an object.

makeev-pavel: After discussion, it was decided to choose such names:
normalizedObjectsHashCache ->
normalizedSpecFragmentsHashCache

componentsHashMap ->
originalSpecComponentsHashCache

src/utils/hashes.ts
makeev-pavel: used _calculateMd5Hash

src/utils/calculateDiffId.ts
makeev-pavel: Added a description

src/builder.ts
makeev-pavel: done

src/utils/hashes.ts
```diff
@@ -17,23 +17,30 @@
...
+}
 
-  // cache by reference for objects and functions
-  //TODO: think of proper way/place to implement caching
-  if (isObject(value)) {
-    const cached = objectHashCache.get(value as object)
-    if (cached !== undefined) {
-      return cached
-    }
+type HashCache = Map<NotUndefined, string> | WeakMap<object, string>
```
b41ex: `HashCache` type seems to be unnecessary anymore, I suggest to use `ObjectHashCache` here.

test/cache.test.ts
```diff
@@ -0,0 +1,73 @@
...
+      expect(calculateObjectHashSpy).toHaveBeenCalledTimes(4)
+
+      const [operation1, operation2] = Array.from(result.operations.values())
+      const deprecatedItems1 = operation1?.deprecatedItems || []
+      const deprecatedItems2 = operation2?.deprecatedItems || []
+      expect(deprecatedItems1[0]?.hash).toBe(deprecatedItems2[0]?.hash)
+    })
+  })
+
+  describe('Unit tests', () => {
```
b41ex: 'Unit tests' -> 'Normalized specification fragments hash cash tests'

b41ex (changes requested):

test/cache.test.ts
makeev-pavel: done

src/utils/hashes.ts
makeev-pavel: done

PR #60, opened 2025-12-08
feat: reclassify diffs to risky in a merged document for no-bwc operation

### LINKED ISSUE #422: [FE] Reclassify diffs to risky in a merged document for no-bwc operation
Walk the merged document.
For operations which are marked as no-bwc, reclassify breaking diffs as follows
- for each such diffs, create a copy, with type changed to risky, 
- remove original diff from DIFF_META_KEY symbol 
- add new diff to DIFF_META_KEY symbol instead
- the same diff could occur several times in the tree and each occurens should be replaced with the same instance of risky diff, so need to keep a map of original diffs to reclassified diffs and not create new if one was already created

Subtrees for operations which are not marked as no-bwc should be skipped.
Should reclassify before diff aggregation.

Add new tests:
- two operations (bwc and no-bwc) use the same schema, breaking change happens in the schema, check that there's both breaking and risky diff for corresponding operations
- the same schema used several times in a no-bwc operation, check that the same risky diff is used in each occurrence

makeev-pavel:
… operation


src/components/compare/bwc.validation.ts
```diff
@@ -14,21 +14,72 @@
...
-      return diff.type === BREAKING_CHANGE_TYPE
-        ? { ...diff, type: RISKY_CHANGE_TYPE }
-        : diff
-    })
+import { isObject, isValidHttpMethod } from '../../utils'
+import { REST_KIND_KEY } from '../../apitypes'
+import { JsonPath } from '@netcracker/qubership-apihub-json-crawl'
+import { ApiCompatibilityKind, BwcState } from '@netcracker/qubership-apihub-api-diff'
+
+const getKind = (obj: unknown): string | undefined => {
```
JayLim2: getExplicitlyDefinedKind

src/components/compare/bwc.validation.ts
```diff
@@ -14,21 +14,72 @@
...
+  return Object.entries(obj).some(([key, value]) =>
+    isValidHttpMethod(key) &&
+    isObject(value) &&
+    hasNoBWC(value),
+  )
+}
+
+const isValidPathLength = (path: JsonPath | undefined): boolean => {
+  const pathLength = path?.length ?? 0
+  return pathLength >= 1 && pathLength <= 3
```
JayLim2: \>= 1 or >= 0?

JayLim2 (approved):

JayLim2 (changes requested):
Some minor changes

src/components/compare/bwc.validation.ts
```diff
@@ -14,21 +14,72 @@
...
+}
+
+export const checkNoBackwardCompatibility = (
```
b41ex: checkApiKind?

src/components/compare/bwc.validation.ts
b41ex: getApiKind?

src/components/compare/bwc.validation.ts
```diff
@@ -14,21 +14,72 @@
...
+  if (hasNoBWC(obj)) return true
+
+  return Object.entries(obj).some(([key, value]) =>
+    isValidHttpMethod(key) &&
+    isObject(value) &&
+    hasNoBWC(value),
+  )
+}
+
+const isValidPathLength = (path: JsonPath | undefined): boolean => {
```
b41ex: 'Valid' is non-descriptive. Use descriptive name like 'isRestOperationLikePath' or something like this

src/components/compare/bwc.validation.ts
b41ex: Add explanatory comment to this magic checks or more granular methods with descriptive names.

src/components/compare/bwc.validation.ts
b41ex: Based on a tests in apiKind.test.ts, no-bwc could be defined in different ways:
- `x-api-kind` extension under specific operation
- `x-api-kind` extension under info object for the whole specification
- via special label during publishing

It seems that current function does not cover latter 2 cases?

Need to improve the tests to not only check for apiKind property value, but also check actual classification of changes in the resulting merged tree and operation summary. And need to fix implement these cases in the function if necessary.

src/components/compare/bwc.validation.ts
```diff
@@ -14,21 +14,72 @@
...
+  }
+
+  if (beforeExists && afterExists) {
+    return hasNoBWC(beforeJson) || hasNoBWC(afterJson)
+      ? ApiCompatibilityKind.NOT_BACKWARD_COMPATIBLE
+      : undefined
+  }
+
+  if (beforeExists && !afterExists) {
+    return hasNoBWCInMethods(beforeJson)
```
b41ex: I am not sure how does it works? We are checking that one of the methods has no-bwc label and if it does, the whole end-point is considered no-bwc? Looks strange to me.

b41ex (changes requested):

src/components/compare/bwc.validation.ts
makeev-pavel: done

src/components/compare/bwc.validation.ts
makeev-pavel: done

src/components/compare/bwc.validation.ts
makeev-pavel: done

src/components/compare/bwc.validation.ts
makeev-pavel: deleted

src/components/compare/bwc.validation.ts
makeev-pavel: done. + added tests

src/components/compare/bwc.validation.ts
makeev-pavel: done

src/components/document.ts
```diff
@@ -77,3 +78,11 @@ const findApiKindLabel = (labels: unknown[]): string => {
   }
   return API_KIND.BWC
 }
+
+export const getApiKind = (obj: unknown, defaultApiKind?: string): string | undefined => {
+  if (!isObject(obj)) {
+    return defaultApiKind
+  }
+  const apiKind = obj?.[REST_KIND_KEY]
+  return typeof apiKind === 'string' ? apiKind?.toLowerCase() : defaultApiKind
```
b41ex: Better to make sure only on of predefined apiKinds (or default) is returned, not arbitrary string.
We already have function which does something similar (rawToApiKind)

src/types/external/documents.ts
```diff
@@ -42,6 +43,7 @@ export type ResolvedVersionDocuments = {
 
 export type ResolvedVersionDocument = ResolvedDocument & {
   packageRef?: string
+  apiKind?: string
```
b41ex: Why ApiKind imported but not used?

src/types/external/version.ts
```diff
@@ -29,7 +30,7 @@ export type ResolvedVersion = {
   operationTypes?: OperationTypes[]
   version: string
   apiProcessorVersion: string
-
+  versionLabels?: Labels
```
b41ex, line 33: As far as I remember when we change resolvers interface, we need to update implementation in `ui` and `build-task-consumer`?

src/components/compare/bwc.validation.ts
```diff
@@ -14,21 +14,144 @@
...
+      isObject(value) &&
+      hasApiKind(value, apiKind),
+    )
+}
+
+const ROOT_PATH_LENGTH = 0
+const PATH_ITEM_PATH_LENGTH = 2
+const MAX_BWC_FLAG_PATH_LENGTH = 3
+
+export const checkApiKind = (
```
JayLim2: createApiKindChecker

src/types/internal/apiBuilder.ts
```diff
@@ -20,7 +20,7 @@ import {
   BuildConfigRef,
   ExportFormat,
   FileId,
-  GroupDocumentsResolver,
+  GroupDocumentsResolver, Labels,
```
JayLim2: new line

src/components/compare/bwc.validation.ts
```diff
@@ -14,21 +14,144 @@
...
+  prevApiKind: string = API_KIND.BWC,
+  currApiKind: string = API_KIND.BWC,
+): ApiCompatibilityScopeFunction => {
+  const defaultApiCompatibilityKind = (prevApiKind === API_KIND.NO_BWC || currApiKind === API_KIND.NO_BWC)
+    ? ApiCompatibilityKind.NOT_BACKWARD_COMPATIBLE
+    : undefined
```
b41ex: Why `undefined` here? How do we ensure api kind `BWC` unless explicitly defined overwise?

src/components/compare/bwc.validation.ts
b41ex: The name of the function is misleading, it does not check api kind, it returns the function.

src/components/compare/bwc.validation.ts
```diff
@@ -14,21 +14,144 @@
...
+  return (
+    path?: JsonPath,
+    beforeJson?: unknown,
+    afterJson?: unknown,
+  ): ApiCompatibilityScope | undefined => {
+    const pathLength = path?.length ?? 0
+    /*
+     * Calculating Api Kind for the entire document as the default
+     * If there is a NO_BWC marker on:
+     * - Publish label
```
b41ex: Let's use exact terms. There are version label and document labels.

src/components/compare/bwc.validation.ts
```diff
@@ -14,21 +14,144 @@
...
+    entries.every(([key, value]) =>
+      isValidHttpMethod(key) &&
+      isObject(value) &&
+      hasApiKind(value, apiKind),
+    )
+}
+
+const ROOT_PATH_LENGTH = 0
+const PATH_ITEM_PATH_LENGTH = 2
+const MAX_BWC_FLAG_PATH_LENGTH = 3
```
b41ex: I would call it `OPERATION_OBJECT_PATH_LENGTH` to be consistent with `PATH_ITEM_PATH_LENGTH`

src/components/compare/bwc.validation.ts
```diff
@@ -14,21 +14,144 @@
...
+    }
+
+    const beforeExists = isObject(beforeJson)
+    const afterExists = isObject(afterJson)
+
+    if (!beforeExists && !afterExists) {
+      return undefined
+    }
+
+    if (beforeExists && afterExists) {
```
b41ex: As far as I understand, this condition is valid only for operation object and following conditions are valid only for path item object. I would explicitly put them under corresponding `if`s to make it clear.

src/components/compare/bwc.validation.ts
```diff
@@ -14,21 +14,144 @@
...
+  return undefined
+}
+
+const hasApiKind = (obj: unknown, apiKind: string): boolean => {
+  return getApiKind(obj) === apiKind
+}
+
+// If a path object is removed/added, we must ensure every HTTP method under it
+// is explicitly marked NO_BWC before treating the change as risky.
+const checkAllMethodsHaveSameApiKind = (obj: unknown, apiKind: string): boolean => {
```
b41ex: This method name is misleading. It is not only checks that all methods have the same kind, but it will also check api kind on the object itself, so, it is used for the particular operation object and it could also behave incorrectly if somebody will put `x-api-kind` extension under path item object (which we do not and should not support).
I suggest to explicitly split logic for path item and operations objects.

src/components/document.ts
```diff
@@ -77,3 +78,11 @@ const findApiKindLabel = (labels: unknown[]): string => {
   }
   return API_KIND.BWC
 }
+
+export const getApiKind = (obj: unknown, defaultApiKind?: string): string | undefined => {
```
b41ex: We should use explicit typing for the return value.

src/components/compare/bwc.validation.ts
```diff
@@ -14,21 +14,144 @@
...
+  ApiCompatibilityKind,
+  ApiCompatibilityScope,
+  ApiCompatibilityScopeFunction,
+} from '@netcracker/qubership-apihub-api-diff'
+import { Labels } from '../../types'
+import { findApiKindLabel, getApiKind } from '../document'
+import { OpenAPIV3 } from 'openapi-types'
+
+export const getApiCompatibilityKind = (beforeJson: unknown, afterJson: unknown): ApiCompatibilityKind | undefined => {
+  const beforeApiKind = getApiKind(beforeJson)?.toLowerCase() ?? ''
```
b41ex: We should not need this `toLowerCase` with proper typing

src/components/compare/bwc.validation.ts
```diff
@@ -14,21 +14,144 @@
...
+    // considered NO_BWC only when every contained method declares NO_BWC.
+    if (afterExists && !beforeExists) {
+      return getMethodsApiCompatibilityKind(afterJson)
+    }
+
+    return undefined
+  }
+}
+
+export const getApiKindFormLabels = (
```
b41ex: typo `getApiKindFormLabels` -> `getApiKindFromLabels`
Also, the name is misleading. Based on its signature, it not only checks labels, but also `info` object.

src/apitypes/rest/rest.changes.ts
```diff
@@ -132,6 +135,10 @@ export const compareDocuments: DocumentsCompare = async (
       normalizedResult: false,
       afterValueNormalizedProperty: AFTER_VALUE_NORMALIZED_PROPERTY,
       beforeValueNormalizedProperty: BEFORE_VALUE_NORMALIZED_PROPERTY,
+      apiCompatibilityScopeFunction: checkApiKind(
+        prevDoc?.apiKind ?? getApiKindFormLabels(prevDocData?.info, prevDoc?.labels, previousVersionLabels),
```
b41ex: I believe it is better to just calculate defaultApiCompatibilityKind here and pass is as a single argument to the `checkApiKind` (to be renamed) function.
So, this piece of code should be somewhere here:
```typescript
const defaultApiCompatibilityKind = (prevApiKind === API_KIND.NO_BWC || currApiKind === API_KIND.NO_BWC)
    ? ApiCompatibilityKind.NOT_BACKWARD_COMPATIBLE
    : undefined
```
Otherwise we have scattered logic, where we calculate api kind for the previous document and for the current document and then combine them later to get default api kind.

src/apitypes/rest/rest.document.ts
```diff
@@ -75,7 +76,7 @@ export const buildRestDocument: DocumentBuilder<OpenAPIV3.Document> = async (par
 
   let bundledFileData = data
 
-  const documentKind = bundledFileData?.info?.[REST_KIND_KEY] || apiKind
+  const documentKind = getApiKind(bundledFileData?.info, apiKind)
```
b41ex: Let's rename it to documentApiKind to avoid confusion.

src/components/compare/bwc.validation.ts
```diff
@@ -14,21 +14,144 @@
...
+import {
+  ApiCompatibilityKind,
+  ApiCompatibilityScope,
+  ApiCompatibilityScopeFunction,
+} from '@netcracker/qubership-apihub-api-diff'
+import { Labels } from '../../types'
+import { findApiKindLabel, getApiKind } from '../document'
+import { OpenAPIV3 } from 'openapi-types'
+
+export const getApiCompatibilityKind = (beforeJson: unknown, afterJson: unknown): ApiCompatibilityKind | undefined => {
```
b41ex: We have `getApiKind` and `getApiCompatibilityKind`. 
From the names it is absolutely not clear what is the difference.

src/components/document.ts
```diff
@@ -48,23 +48,24 @@ export const buildDocument = async (parsedFile: SourceFile, file: BuildConfigFil
...
 
     return await apiBuilder.buildDocument(parsedFile, file, ctx)
   } catch (error) {
     throw new Error(`Cannot process the "${file.fileId}" document. ${error instanceof Error ? error.message : 'Unknown error'}`)
   }
 }
 
-const findApiKindLabel = (labels: unknown[]): string => {
-  if (!Array.isArray(labels)) {
+export const findApiKindLabel = (fileLabels: unknown, versionLabels: unknown): string => {
```
b41ex: I would call it `calculateApiKindFromLabels`

test/apiKinds.test.ts
```diff
@@ -156,3 +166,196 @@ describe('Risky changes for no-bwc operations test', () => {
...
+  const CURR_VERSION = 'v2'
+  const API_KIND_NO_BWC_LABEL = 'apihub/x-api-kind: no-BWC'
+  const API_KIND_BWC_LABEL = 'apihub/x-api-kind: BWC'
+
+  describe('Publish ApiKind version Labels tests', () => {
+    test('should apply api kind version label from previous document', async () => {
+      const result = await runApiKindTest(
+        'api-kinds/no-api-kind-in-documents', [], [], [API_KIND_NO_BWC_LABEL],
+      )
+      expect(result).toEqual(changesSummaryMatcher({ [RISKY_CHANGE_TYPE]: 1 }))
```
b41ex: I think we should check the actual diff in the version comparison document. 
This check would pass with the old code, so it is not checking the behavior we need.

test/apiKinds.test.ts
```diff
@@ -156,3 +166,196 @@ describe('Risky changes for no-bwc operations test', () => {
...
 })
+
+describe('Check Api Compatibility Function tests', () => {
+  const PREV_VERSION = 'v1'
+  const CURR_VERSION = 'v2'
+  const API_KIND_NO_BWC_LABEL = 'apihub/x-api-kind: no-BWC'
+  const API_KIND_BWC_LABEL = 'apihub/x-api-kind: BWC'
+
+  describe('Publish ApiKind version Labels tests', () => {
+    test('should apply api kind version label from previous document', async () => {
```
b41ex: ".. from previous version", not "... from previous document"

test/apiKinds.test.ts
```diff
@@ -156,3 +166,196 @@ describe('Risky changes for no-bwc operations test', () => {
...
+      const result = await runApiKindTest(
+        'api-kinds/no-api-kind-in-documents', [], [], [API_KIND_NO_BWC_LABEL],
+      )
+      expect(result).toEqual(changesSummaryMatcher({ [RISKY_CHANGE_TYPE]: 1 }))
+    })
+
+    test('should apply api kind version label from current document', async () => {
```
b41ex: .. from current version

test/apiKinds.test.ts
```diff
@@ -156,3 +166,196 @@ describe('Risky changes for no-bwc operations test', () => {
...
+
+    test('should apply api kind label from current document', async () => {
+      const result = await runApiKindTest('api-kinds/no-api-kind-in-documents', [], [API_KIND_NO_BWC_LABEL])
+      expect(result).toEqual(changesSummaryMatcher({ [RISKY_CHANGE_TYPE]: 1 }))
+    })
+  })
+
+  describe('ApiKind info section tests', () => {
+    test('should apply api-kind label in previous document info section', async () => {
+      const result = await runApiKindTest('api-kinds/no-bwc-api-kind-info-label-in-prev-document')
```
b41ex: In `info` section we have `x-api-kind` **property** or **extension**, not label. 
Let's not confuse terms.

test/apiKinds.test.ts
```diff
@@ -156,3 +166,196 @@ describe('Risky changes for no-bwc operations test', () => {
...
+      const result = await runApiKindTest('api-kinds/no-bwc-api-kind-operation-label-prev-document')
+      expect(result).toEqual(changesSummaryMatcher({
+        [RISKY_CHANGE_TYPE]: 1,
+        [UNCLASSIFIED_CHANGE_TYPE]: 1, // x-api-kind field in the operation also gives a diff
+      }))
+    })
+  })
+
+  describe('Publish ApiKind Priority tests', () => {
+    test('should prioritize version label no-BWC api kind over file label BWC in previous version', async () => {
```
b41ex: The name is wrong, judging from ER it is vice versa

test/apiKinds.test.ts
```diff
@@ -156,3 +166,196 @@ describe('Risky changes for no-bwc operations test', () => {
...
+      const result = await runApiKindTest(
+        'api-kinds/no-api-kind-in-documents', [API_KIND_BWC_LABEL], [], [API_KIND_NO_BWC_LABEL],
+      )
+      expect(result).toEqual(changesSummaryMatcher({ [BREAKING_CHANGE_TYPE]: 1 }))
+    })
+
+    test('should prioritize version label no-BWC api kind over file label BWC in current version', async () => {
```
b41ex: The name is wrong, judging from ER it is vice versa

test/apiKinds.test.ts
```diff
@@ -156,3 +166,196 @@ describe('Risky changes for no-bwc operations test', () => {
...
+    })
+
+    test('should prioritize file label no-BWC api kind over version label BWC in current version', async () => {
+      const result = await runApiKindTest(
+        'api-kinds/no-api-kind-in-documents', [], [API_KIND_NO_BWC_LABEL], [], [API_KIND_BWC_LABEL],
+      )
+      expect(result).toEqual(changesSummaryMatcher({ [RISKY_CHANGE_TYPE]: 1 }))
+    })
+
+    test('should prioritize info label no-BWC api kind over version label BWC in previous version', async () => {
```
b41ex: It is actually prioritizing over file label. I think setting version labels is excessive in this case, given the other cases.
Also- info property/extension, not label.

test/apiKinds.test.ts
```diff
@@ -156,3 +166,196 @@ describe('Risky changes for no-bwc operations test', () => {
...
+      const result = await runApiKindTest(
+        'api-kinds/no-bwc-api-kind-info-label-in-prev-document', [API_KIND_BWC_LABEL], [], [API_KIND_BWC_LABEL],
+      )
+      expect(result).toEqual(changesSummaryMatcher({ [RISKY_CHANGE_TYPE]: 1 }))
+    })
+
+    test('should prioritize info label no-BWC api kind over version label BWC in current version', async () => {
```
b41ex: Same

test/apiKinds.test.ts
```diff
@@ -156,3 +166,196 @@ describe('Risky changes for no-bwc operations test', () => {
...
+    })
+
+    test('should prioritize current info label as BWC when info api kind overrides no-BWC publish labels', async () => {
+      const result = await runApiKindTest(
+        'api-kinds/bwc-api-kind-info-label-in-curr-document', [], [API_KIND_NO_BWC_LABEL], [], [API_KIND_NO_BWC_LABEL],
+      )
+      expect(result).toEqual(changesSummaryMatcher({ [BREAKING_CHANGE_TYPE]: 1 }))
+    })
+
+    test('should prioritize previous operation-level BWC api kind over current no-BWC publish labels', async () => {
```
b41ex: The current behavior in `main` seem to be different- it is enough to have compatibility state calculated to `no-BWC` for one of the versions to override `BWC` set for the other version.

Need to check current behavior and design and align.

test/projects/api-kinds/operation-bwc-in-curr-document/1.yaml
b41ex, line 1: It is better to explicitly have values in the name, not just `double`. Something like `info-bwc-operation-noBWC-in-curr-document`

b41ex (changes requested):

src/components/compare/bwc.validation.ts
makeev-pavel: I see different suggestions by name. It is necessary to coordinate

src/types/external/documents.ts
makeev-pavel: we use
https://github.com/Netcracker/qubership-apihub-api-processor/pull/60/files#diff-99318dedcbc933c30fc59b9467ecda15bb786d0b0589deeacbe7796ca2fc0679R139

Also versionDocumentsResolver add apikind to ResolvedVersionDocument

src/types/external/version.ts
makeev-pavel: I don't understand yet what impact these fields can have. We also have [key: string]: unknown

src/components/compare/bwc.validation.ts
makeev-pavel: done

src/types/internal/apiBuilder.ts
makeev-pavel: done

src/components/compare/bwc.validation.ts
makeev-pavel: I know that there is a default value in api-diff and it will be taken. But it's not obvious here. Fixed it

src/components/compare/bwc.validation.ts
makeev-pavel: createApiKindChecker

src/components/compare/bwc.validation.ts
makeev-pavel: done

src/components/compare/bwc.validation.ts
makeev-pavel: done

src/components/compare/bwc.validation.ts
makeev-pavel: done

src/components/compare/bwc.validation.ts
makeev-pavel: I have removed the field check of the object itself.

src/components/compare/bwc.validation.ts
makeev-pavel: getApiKindFromLabels - ok
I assumed that the info section also had a kind of label. Ready to discuss the name

src/apitypes/rest/rest.document.ts
makeev-pavel: done

src/components/document.ts
makeev-pavel: done

test/apiKinds.test.ts
makeev-pavel: The old tests have a large file and a lot of changes. 
These tests are tailored to specific functionality. There are one or two changes in them.  To describe the changes is to complicate the logic and understanding of what the test is doing

test/apiKinds.test.ts
makeev-pavel: Corrected the name. Based on the documentation about version priorities, we need to check this case.

test/projects/api-kinds/operation-bwc-in-curr-document/1.yaml
makeev-pavel: done

test/apiKinds.test.ts
makeev-pavel: fixed labels

test/apiKinds.test.ts
makeev-pavel: done

test/apiKinds.test.ts
makeev-pavel: done

test/apiKinds.test.ts
makeev-pavel: done

test/apiKinds.test.ts
makeev-pavel: done

test/apiKinds.test.ts
makeev-pavel: done

test/apiKinds.test.ts
makeev-pavel: done

src/components/compare/bwc.validation.ts
makeev-pavel: ApiKind from api-processor
ApiCompatibilityKind - from api-diff

src/apitypes/rest/rest.changes.ts
makeev-pavel: We have priorities. Labels on different levels and info field. I've put the logic for calculating the final ApiKind in a separate function. It seems to me that this is not the place for these calculations.

src/components/document.ts
makeev-pavel: done

src/types/external/documents.ts
makeev-pavel: done

src/types/external/version.ts
makeev-pavel: Ui + build-task-consumer  checked

src/components/compare/bwc.validation.ts
makeev-pavel: the place of verification has changed slightly

src/components/document.ts
makeev-pavel: done

src/components/compare/bwc.validation.ts
makeev-pavel: done

src/components/compare/bwc.validation.ts
makeev-pavel: done

src/apitypes/rest/rest.changes.ts
makeev-pavel: done

src/apitypes/graphql/graphql.operation.ts
```diff
@@ -85,13 +85,11 @@ export const buildGraphQLOperation = (
...
   }, debugCtx)
 
-  const apiKind = documentApiKind || API_KIND.BWC
-
   return {
     operationId,
     documentId: documentSlug,
     apiType: GRAPHQL_API_TYPE,
-    apiKind: rawToApiKind(apiKind),
+    apiKind: rawToApiKind(documentApiKind, APIHUB_API_COMPATIBILITY_KIND.BWC),
```
b41ex: It is strange to see call to `rawToApiKind` when we are taking some field from structured object. 
It is already of the right type, there should be no need to transform it from raw, it is already done in `document.ts` buildDocument:


```typescript
file.apiKind = calculateApiKindFromLabels(file.labels, ctx.versionLabels)
```

Btw, default is also applied there.

src/apitypes/rest/rest.changes.ts
```diff
@@ -121,6 +125,9 @@ export const compareDocuments: DocumentsCompare = async (
     currDocData = createCopyWithEmptyPathItems(prevDocData)
   }
 
+  const prevApiKind = prevDoc?.apiKind || getApiKind(prevDocData?.info) || calculateApiKindFromLabels(prevDoc?.labels, previousVersionLabels)
```
b41ex: As far as I remember, property specified in `info` should be of higher priority than from labels (and hence higher than `prevDoc?.apiKind` which is calculated from labels. 
Btw, why the need to calculate apiKind for the document again when we already calculated it in `document.ts` buildDocument? And priority of api kind property from info is ensured in `rest.document.ts` ?

What cases fail if you just remove these 2 lines and take apiKind from prevDoc and currDoc?

src/components/document.ts
```diff
@@ -48,32 +55,43 @@ export const buildDocument = async (parsedFile: SourceFile, file: BuildConfigFil
...
     const match = new RegExp(`${API_KIND_LABEL}:`).exec(label)
     if (match) {
-      return rawToApiKind(label.slice(match[0].length).trim())
+      return rawToApiKind(label.slice(match[0].length).trim(), APIHUB_API_COMPATIBILITY_KIND.BWC)
+    }
+  }
+  return APIHUB_API_COMPATIBILITY_KIND.BWC
+}
+
+export const getApiKind = (obj: unknown, defaultApiKind?: ApihubApiCompatibilityKind): ApihubApiCompatibilityKind | undefined => {
```
b41ex: I propose to rename it to `getApiKindProperty`

src/components/compare/bwc.validation.ts
```diff
@@ -14,21 +14,142 @@
...
+      isObject(value) &&
+      hasApiKind(value, apiKind),
+    )
+}
+
+const ROOT_PATH_LENGTH = 0
+const PATH_ITEM_PATH_LENGTH = 2
+const OPERATION_OBJECT_PATH_LENGTH = 3
+
+export const createApiKindChecker = (
```
b41ex: `createApihubApiCompatibilityScopeFunction` ?

src/apitypes/rest/rest.operation.ts
```diff
@@ -142,7 +142,7 @@ export const buildRestOperation = (
   }, debugCtx)
 
   const models: Record<string, string> = {}
-  const apiKind = effectiveOperationObject[REST_KIND_KEY] || documentApiKind || API_KIND.BWC
+  const apiKind = effectiveOperationObject[REST_KIND_KEY] || documentApiKind || APIHUB_API_COMPATIBILITY_KIND.BWC
```
b41ex: 1. Let call the variable operationApiKind for easier understanding of what we are calculating here.
2. `effectiveOperationObject[REST_KIND_KEY]`- you have introduced a function just for it `getApiKind` (`getApiKindProperty`), lets used it

src/apitypes/rest/rest.operation.ts
```diff
@@ -171,7 +171,7 @@ export const buildRestOperation = (
     operationId,
     documentId: documentSlug,
     apiType: REST_API_TYPE,
-    apiKind: rawToApiKind(apiKind),
+    apiKind: rawToApiKind(apiKind, APIHUB_API_COMPATIBILITY_KIND.BWC),
```
b41ex: `rawToApiKind` should be applied when we are converting data from raw text to some typed variable, it should have happened by now.

src/components/compare/bwc.validation.ts
```diff
@@ -14,21 +14,142 @@
...
+  beforeJson: unknown,
+  afterJson: unknown,
+  beforeParentApiKind: ApihubApiCompatibilityKind,
+  afterParentApiKind: ApihubApiCompatibilityKind,
+): ApiCompatibilityKind | undefined => {
+  const beforeKind = getApiKind(beforeJson) ?? beforeParentApiKind
+  const afterKind = getApiKind(afterJson) ?? afterParentApiKind
+
+  if (beforeKind === APIHUB_API_COMPATIBILITY_KIND.NO_BWC || afterKind === APIHUB_API_COMPATIBILITY_KIND.NO_BWC) {
+    return convertApihubToApiCompatibilityKind(APIHUB_API_COMPATIBILITY_KIND.NO_BWC)
```
b41ex: `convertApihubToApiCompatibilityKind` are used just for constant, why not just return corresponding constant from api-diff type? If think given its usages `convertApihubToApiCompatibilityKind` is redundant.

src/components/compare/bwc.validation.ts
```diff
@@ -14,21 +14,142 @@
...
+    case APIHUB_API_COMPATIBILITY_KIND.BWC:
+      return ApiCompatibilityKind.BACKWARD_COMPATIBLE
+    case APIHUB_API_COMPATIBILITY_KIND.NO_BWC:
+      return ApiCompatibilityKind.NOT_BACKWARD_COMPATIBLE
+    default:
+      return ApiCompatibilityKind.BACKWARD_COMPATIBLE
+  }
+}
+
+export const getApiCompatibilityKind = (
```
b41ex: `calculateOperationApiCompatibilityKind`
And since we know that this method could be called only for operation object, we can use appropriate type for beforeJson/afterJson instead of unknown.

src/components/compare/bwc.validation.ts
```diff
@@ -14,21 +14,142 @@
...
+  beforeJson: unknown,
+  afterJson: unknown,
+  beforeParentApiKind: ApihubApiCompatibilityKind,
```
b41ex: From the name `beforeParentApiKind` it seems that it could change from parent to parent as we walk the tree. When in fact with the current logic it is just beforeDefaulApiKind and afterDefaultApiKind. I think it is better call them that to avoid unnecessary confusion.

src/apitypes/rest/rest.changes.ts
```diff
@@ -121,6 +125,9 @@ export const compareDocuments: DocumentsCompare = async (
...
+  const currApiKind = currDoc?.apiKind || getApiKind(currDocData?.info) || calculateApiKindFromLabels(currDoc?.labels, currentVersionLabels)
```
b41ex: If we do need this variables- let's call the `prevDocumentApiKind` and `currDocumentApiKind`

src/components/compare/bwc.validation.ts
```diff
@@ -14,21 +14,142 @@
...
+import { JsonPath } from '@netcracker/qubership-apihub-json-crawl'
+import {
+  ApiCompatibilityKind,
+  ApiCompatibilityScope,
+  ApiCompatibilityScopeFunction,
+} from '@netcracker/qubership-apihub-api-diff'
+import { ApihubApiCompatibilityKind } from '../../types'
+import { getApiKind } from '../document'
+
+const convertApihubToApiCompatibilityKind = (apihubApiKind: ApihubApiCompatibilityKind): ApiCompatibilityKind => {
```
b41ex: Redundant function with the current usage- just return constants of appropriate type in each place where it is used.

src/components/compare/bwc.validation.ts
```diff
@@ -14,21 +14,142 @@
...
+
+  if (beforeKind === APIHUB_API_COMPATIBILITY_KIND.NO_BWC || afterKind === APIHUB_API_COMPATIBILITY_KIND.NO_BWC) {
+    return convertApihubToApiCompatibilityKind(APIHUB_API_COMPATIBILITY_KIND.NO_BWC)
+  }
+
+  if (beforeKind === APIHUB_API_COMPATIBILITY_KIND.BWC && afterKind === APIHUB_API_COMPATIBILITY_KIND.BWC) {
+    return convertApihubToApiCompatibilityKind(APIHUB_API_COMPATIBILITY_KIND.BWC)
+  }
+
+  return undefined
```
b41ex: Couldn't happen with the current  logic?

src/components/compare/bwc.validation.ts
```diff
@@ -14,21 +14,142 @@
...
+}
+
+const hasApiKind = (obj: unknown, apiKind: string): boolean => {
+  return getApiKind(obj) === apiKind
+}
+
+// If a path object is removed/added, we must ensure every HTTP method under it
+// is explicitly marked NO_BWC before treating the change as risky.
+const checkAllMethodsHaveSameApiKind = (obj: unknown, apiKind: string): boolean => {
```
b41ex: Since this method could be called only for PathItemObject, we might as well use exact type instead of `unknown`

src/components/compare/bwc.validation.ts
```diff
@@ -14,21 +14,142 @@
...
+      hasApiKind(value, apiKind),
+    )
+}
+
+const ROOT_PATH_LENGTH = 0
+const PATH_ITEM_PATH_LENGTH = 2
+const OPERATION_OBJECT_PATH_LENGTH = 3
+
+export const createApiKindChecker = (
+  prevApiKind: ApihubApiCompatibilityKind = APIHUB_API_COMPATIBILITY_KIND.BWC,
```
b41ex: `prevDocumentApiKind`?

src/types/external/operations.ts
```diff
@@ -16,10 +16,10 @@
 
 import { KeyOfConstType, OperationId, OperationsApiType, PackageId, VersionId } from './types'
 import { DeprecateItem } from './deprecated'
-import { API_KIND } from '../../consts'
+import { APIHUB_API_COMPATIBILITY_KIND } from '../../consts'
 import { ApiAudience } from '../package'
 
-export type ApiKind = KeyOfConstType<typeof API_KIND>
+export type ApihubApiCompatibilityKind = KeyOfConstType<typeof APIHUB_API_COMPATIBILITY_KIND>
```
b41ex: Let's use union of constants like [here](https://github.com/Netcracker/qubership-apihub-class-view/blob/7d02e45e7b913db44cb3830f54efd18e28b6d3b5/src/main/domain/like/type.ts#L17). 
I remember we discussed it some time ago and decided this is the approach we would follow.

And in any case I believe it is better to have declaration of APIHUB_API_COMPATIBILITY_KIND and ApihubApiCompatibilityKind in single place, not separate files.

test/helpers/matchers.ts
```diff
@@ -117,6 +124,30 @@ export function operationTypeMatcher(
...
+          comparisonInternalDocuments: expect.arrayContaining([
+            expect.objectContaining(expected),
+          ]),
+        }),
+      ]),
+    },
+  )
+}
+
+export function serializedComparisonDocumentMatcher(
```
b41ex: This is too broad/fragile matcher. We are essentially checking for substring breaking/risky in a file and it could easily fail to fail even if be do break something. I believe we need to be more precise in checking (by particular path).

test/apiKinds.test.ts
```diff
@@ -156,3 +166,354 @@ describe('Risky changes for no-bwc operations test', () => {
...
+      })
+    })
+  })
+
+  describe('ApiKind operations section tests', () => {
+    test('should apply operation no-BWC in current document', async () => {
+      const result = await runApiKindTest('api-kinds/operation-noBWC-in-curr-document')
+      expect(result).toEqual(changesSummaryMatcher({
+        [RISKY_CHANGE_TYPE]: 1,
+        [UNCLASSIFIED_CHANGE_TYPE]: 1, // x-api-kind field in the operation also gives a diff
```
b41ex: Could put it into constant with appropriate name (like `x-api-kind-change` and spread it everywhere you need it).
Will allow to not add comment in every place and also if we revise classification of extension change in the future- it will be easier to change expected result in one place.

src/components/compare/bwc.validation.ts
```diff
@@ -14,21 +14,142 @@
...
+      return undefined
+    }
+    const beforeExists = isObject(beforeJson)
+    const afterExists = isObject(afterJson)
+
+    if (!beforeExists && !afterExists) {
+      return undefined
+    }
+
+    if (pathLength === PATH_ITEM_PATH_LENGTH) {
```
b41ex, line 134: I believe we are missing tests for path item logic bwc check logic in api-processor.

b41ex (changes requested):

src/components/compare/bwc.validation.ts
```diff
@@ -14,21 +14,142 @@
...
+// If a path object is removed/added, we must ensure every HTTP method under it
+// is explicitly marked NO_BWC before treating the change as risky.
+const checkAllMethodsHaveSameApiKind = (obj: unknown, apiKind: string): boolean => {
+  if (!isObject(obj)) {
+    return false
+  }
+  const entries = Object.entries(obj)
+
+  return entries.length > 0 &&
+    entries.every(([key, value]) =>
```
b41ex: Path Item Object could have specification extensions under it, which are not Operation Objects. I believe this method will work incorrectly in this case.

src/components/compare/bwc.validation.ts
```diff
@@ -14,21 +14,142 @@
...
+
+      // case add: additions are checked the same way. A new path or operation is
+      // considered NO_BWC only when every contained method declares NO_BWC.
+      if (afterExists && !beforeExists) {
+        return getMethodsApiCompatibilityKind(afterJson)
+      }
+    }
+
+    if (pathLength === OPERATION_OBJECT_PATH_LENGTH) {
+      return getApiCompatibilityKind(beforeJson, afterJson, prevApiKind, currApiKind)
```
b41ex: Again, may be it is worth considering that we could have OAS extension here as well as operation object.

b41ex (changes requested):

src/apitypes/rest/rest.changes.ts
makeev-pavel: In the current document, we should have the kind api already calculated.
It is calculated in the same way by priority
```
file.apiKind = calculateApiKindFromLabels(file.labels, ctx.versionLabels)
const documentApiKind = getApiKindProperty(bundledFileData?.info) || apiKind
```
We don't have an ApiKind for the previous document. And you need to calculate it yourself through info and labels.
Did I do the same 
`prevDoc calculation?.apiKind || getApiKind(prevDocData?.info ) || calculateApiKindFromLabels(prevDoc?.labels, previousVersionLabels)`
for prev and curr - to make it look beautiful and not misleading. We'll get ApiKind in the simplest way if possible.

src/components/compare/bwc.validation.ts
makeev-pavel: Tell me more precisely when isValidHttpMethod is not enough for us?

src/apitypes/graphql/graphql.operation.ts
makeev-pavel: yes, before that it was considered a link. after correcting the types, this is no longer necessary. fixed

src/components/document.ts
makeev-pavel: done

src/components/compare/bwc.validation.ts
makeev-pavel: We need a discussion

src/apitypes/rest/rest.operation.ts

src/apitypes/rest/rest.operation.ts
makeev-pavel: yes, before that it was considered a link. after correcting the types, this is no longer necessary. fixed

src/components/compare/bwc.validation.ts
makeev-pavel: I think we discussed the need for explicit transformation of different types in a separate function.

src/components/compare/bwc.validation.ts
makeev-pavel: done

src/components/compare/bwc.validation.ts
makeev-pavel: done

src/apitypes/rest/rest.changes.ts
makeev-pavel: need. done

src/components/compare/bwc.validation.ts
makeev-pavel: I think we discussed the need for explicit transformation of different types in a separate function.

src/components/compare/bwc.validation.ts
makeev-pavel: these are not all available ways. We have 'experimental' ApiKind. I think it is necessary to return BWC in such cases.

src/components/compare/bwc.validation.ts
makeev-pavel: done

src/components/compare/bwc.validation.ts
makeev-pavel: done

src/types/external/operations.ts
makeev-pavel: done

test/helpers/matchers.ts
makeev-pavel: We need to discuss it. I don't like the idea of searching for fields in the generated flatted string. If flatted is updated and the logic changes there, we will have to rewrite the tests. It may be necessary to deserialize the document at the beginning

test/apiKinds.test.ts
makeev-pavel: this is specific only to operations. I added OPERATION_RISKY_CHANGE_TYPES and OPERATION_BREAKING_CHANGE_TYPES

src/components/compare/bwc.validation.ts
makeev-pavel: added

src/apitypes/graphql/graphql.operation.ts
```diff
@@ -85,13 +84,11 @@ export const buildGraphQLOperation = (
...
   }, debugCtx)
 
-  const apiKind = documentApiKind || API_KIND.BWC
-
   return {
     operationId,
     documentId: documentSlug,
     apiType: GRAPHQL_API_TYPE,
-    apiKind: rawToApiKind(apiKind),
+    apiKind: documentApiKind || APIHUB_API_BWC_KIND,
```
b41ex: Can we avoid spreading logic for default value of API kind and set it in one place?

src/apitypes/rest/rest.changes.ts
b41ex: Actually I think calculating currDocumentApiKind in a more complex way than it is necessary looks misleading. 
I would propose to just take already calculated value for `currDocumentApiKind` and for `prevDocumentApiKind` leave more complex calculating, but add comment explaining why it is required.

src/consts.ts
```diff
@@ -108,11 +108,13 @@ export const VERSION_STATUS = {
   NONE: '', // non-existent status for changelog builds
 } as const
 
-export const API_KIND = {
-  BWC: 'bwc',
-  NO_BWC: 'no-bwc',
-  EXPERIMENTAL: 'experimental',
-} as const
+export const APIHUB_API_BWC_KIND = 'bwc'
```
b41ex: > Names of types or constants related to the same entity should have the common part at the beginning. E.g., SUB_KEY_TITLE, SUB_KEY_INNER_LINK

Hence:
APIHUB_API_COMPATIBILITY_KIND_BWC 
APIHUB_API_COMPATIBILITY_KIND_NO_BWC 
APIHUB_API_COMPATIBILITY_KIND_EXPERIMENTAL

test/helpers/matchers.ts
```diff
@@ -231,3 +256,37 @@ export function securitySchemesFromRequirementsMatcher(
...
+
+  return comparisonDocumentMatcher({
+    serializedComparisonDocument: {
+      asymmetricMatch: (actual: string): boolean => {
+        try {
+          const deserializedDoc = deserializeDocument(actual)
+          const diffs = extractAllDiffsFromDocument(deserializedDoc)
+
+          const diffTypes = new Set(diffs.map(diff => diff.type))
+          return apiKinds.every(apiKind => diffTypes.has(apiKind))
```
b41ex, line 286: This matcher turned out to be quite specialized (only applicable to the test data where exactly one diff is expected).
The better version would have similar contract to `diffsMatcher` from api-diff, allow to change a set of diffs in the comparisonDocument.
Can we reuse `diffsMatcher` from api-diff here?
And I would propose to call it `comparisonDocumentDiffMatcher`, otherwise it is not clear what exactly it is checking in the comparisonDocument.

src/components/compare/bwc.validation.ts
b41ex: I don't get how extensions are handled in the current logic. I see only casting to `OpenAPIV3.OperationObject` (which will be incorrect if OAS extensions will be added/removed under path item)

test/helpers/utils.ts
```diff
@@ -266,3 +272,15 @@ export async function prepareChangelogDashboard(
     status: VERSION_STATUS.RELEASE,
   })
 }
+
+const DESERIALIZE_SYMBOL_STRING_MAPPING = new Map([
```
b41ex: Why create separate map, when it should be exactly the same as existing map (`SERIALIZE_SYMBOL_STRING_MAPPING`.) used during serialization to work correctly? Just use `SERIALIZE_SYMBOL_STRING_MAPPING` in deserialize

test/helpers/utils.ts
```diff
@@ -266,3 +272,15 @@ export async function prepareChangelogDashboard(
...
+  ['HASH_FLAG', HASH_FLAG],
+  ['ORIGINS_SYMBOL', ORIGINS_SYMBOL],
+  ['SYNTHETIC_TITLE_FLAG', SYNTHETIC_TITLE_FLAG],
+  ['DIFF_META_KEY', DIFF_META_KEY],
+  ['DIFFS_AGGREGATED_META_KEY', DIFFS_AGGREGATED_META_KEY],
+])
+
+export function deserializeDocument(normalizedDocument: string): ApiDocument {
```
b41ex: `normalizedDocument` -> `serializedDocument`

src/apitypes/rest/rest.changes.ts
b41ex: Actually I think calculating currDocumentApiKind in a more complex way than it is necessary looks misleading. 
I would propose to just take already calculated value for `currDocumentApiKind` and for `prevDocumentApiKind` leave more complex calculating, but add comment explaining why it is required.

b41ex (changes requested):

src/components/compare/bwc.validation.ts
makeev-pavel: added Remove operations/pathItem tests

b41ex: Let's split 'Remove operations/pathItem tests' test group in 2 groups- one where operations are removed, another with tests where entire path item is removed.

test/helpers/utils.ts
makeev-pavel: done

test/helpers/utils.ts
makeev-pavel: fixed to 
`const DESERIALIZE_SYMBOL_STRING_MAPPING = invertMap(SERIALIZE_SYMBOL_STRING_MAPPING)`

src/consts.ts
makeev-pavel: done

src/apitypes/graphql/graphql.operation.ts
makeev-pavel: Now, when creating a VersionDocument, we set the required apiKind. Getting it from a file, or by priority (as in rest). For those places where apiKind was not specified, the default BWC is used.

makeev-pavel:
currDocumentApiKind - fixed. We get it from the document. Tested
`pathLength === OPERATION_OBJECT_PATH_LENGTH ` - we already have tests for this. I divided the test groups into PathItem removal and operations removal. 
Also all ApiKind tests tested on 4.2.4 tag

test/helpers/matchers.ts
makeev-pavel: A lot of time has been spent, but it has not yet been possible to use diffsMatcher. Several LLMs also didn't help, the expected result doesn't correct. I suggest we leave this as a non-priority remark.

b41ex: Let's add TODO

src/apitypes/rest/rest.operation.ts
```diff
@@ -172,7 +170,7 @@ export const buildRestOperation = (
     operationId,
     documentId: documentSlug,
     apiType: REST_API_TYPE,
-    apiKind: operationApiKind,
+    apiKind: documentApiKind,
```
b41ex: Are you sure this is correct replacement?
What if operation api kind is different from the document api kind?
I don't understand why no tests failed for this change.

src/apitypes/unknown/unknown.document.ts
```diff
@@ -59,12 +59,13 @@ export const buildUnknownDocument: DocumentBuilder<string> = async (parsedFile,
     metadata,
     errors: parsedFile.errors?.length ?? 0,
     source,
+    apiKind: apiKindFormLabels || APIHUB_API_COMPATIBILITY_KIND_BWC,
```
b41ex: For unknown document and binary document there's no notion of api kind.
If we can not make apiKind mandatory without introducing them for unknow and binary document- let's revert this change.

src/apitypes/unknown/unknown.document.ts
```diff
@@ -81,6 +82,7 @@ export const buildBinaryDocument: (parsedFile: SourceFile, file: BuildConfigFile
     filename: fileId,
     metadata,
     source,
+    apiKind: apiKindFormLabels || APIHUB_API_COMPATIBILITY_KIND_BWC,
```
b41ex: Ditto

src/components/document.ts
```diff
@@ -48,6 +48,7 @@ export const buildErrorDocument = (file: BuildConfigFile, parsedFile?: TextFile)
     operationIds: [],
     metadata,
     source: parsedFile?.source,
+    apiKind: apiKindFromLabels || APIHUB_API_COMPATIBILITY_KIND_BWC,
```
b41ex: Ditto

src/strategies/merged-document-group.strategy.ts
```diff
@@ -87,6 +87,7 @@ export class MergedDocumentGroupStrategy implements BuilderStrategy {
       metadata: {},
       publish: true,
       versionInternalDocument: createVersionInternalDocument(title),
+      apiKind: APIHUB_API_COMPATIBILITY_KIND_BWC,
```
b41ex: Ditto

test/helpers/utils.ts
```diff
@@ -273,14 +270,14 @@ export async function prepareChangelogDashboard(
...
 }
 
-const DESERIALIZE_SYMBOL_STRING_MAPPING = new Map([
-  ['HASH_FLAG', HASH_FLAG],
-  ['ORIGINS_SYMBOL', ORIGINS_SYMBOL],
-  ['SYNTHETIC_TITLE_FLAG', SYNTHETIC_TITLE_FLAG],
-  ['DIFF_META_KEY', DIFF_META_KEY],
-  ['DIFFS_AGGREGATED_META_KEY', DIFFS_AGGREGATED_META_KEY],
-])
+const invertMap = (map: Map<symbol, string>): Map<string, symbol> => {
```
b41ex: This method does not use type definitions (symbol, string) in its logic. 
I would propose to make it as generic as possible (if we want to define typing- let's define the loosest possible types required for the method logic itself.

b41ex (changes requested):

test/helpers/utils.ts
makeev-pavel: done

src/strategies/merged-document-group.strategy.ts
makeev-pavel: revert

src/components/document.ts
makeev-pavel: revert

src/apitypes/unknown/unknown.document.ts
makeev-pavel: revert

src/apitypes/unknown/unknown.document.ts
makeev-pavel: revert

src/apitypes/rest/rest.operation.ts
makeev-pavel: revert

test/helpers/matchers.ts
makeev-pavel: added

b41ex (approved):

PR #70, opened 2026-03-13
feat: build changelog for AsyncAPI specifications

### LINKED ISSUE #513: [FE] Build changelog for AsyncAPI specifications
Implement build type `changelog` in api-processor for AsyncAPI specifications.

The approach is the same as for REST and GraphQL- process whole specification(s) using api-diff, collect data for specific APIHUB operations from merged tree.

Diffs for corresponding message should be collected for sure. Need to define which other parts of the specification should contribute to the single APIHUB operation diffs (channel, (asyncapi) operation, some other parts of the specification related to the APIHUB operation?)

makeev-pavel:


src/apitypes/async/async.changes.ts
```diff
@@ -101,6 +103,49 @@ export const compareDocuments: DocumentsCompare = async (
...
+      const operation = operationsMap[key]
+      return operation?.previous?.metadata?.asyncOperationId === asyncOperationId || operation?.current?.metadata?.asyncOperationId === asyncOperationId
+    })
+    const operation = matchingOperations.find(matchingOperation => matchingOperation.endsWith(String(index + 1)))
+
+    return operation || matchingOperations[index]
+  }
+
   // Iterate through operations in merged document
   const { operations } = merged
```
b41ex: Let us be consistent in naming.
Since `operation` is used extensively in APIHUB code to designate APIHUB operation, let us explicitly call operations in AsyncAPI term `asyncOperation`

src/apitypes/async/async.changes.ts
```diff
@@ -137,18 +183,23 @@ export const compareDocuments: DocumentsCompare = async (
...
+          const allOperationDiffs = (operationObject as WithAggregatedDiffs<AsyncAPIV3.OperationObject>)[DIFFS_AGGREGATED_META_KEY] ?? []
+
+          const otherMessageDiffs = collectOtherMessageDiffs(messages, messageIndex)
           operationDiffs = [
-            ...(operationObject as WithAggregatedDiffs<AsyncAPIV3.OperationObject>)[DIFFS_AGGREGATED_META_KEY] ?? [],
-            // TODO: check
-            // ...extractAsyncApiVersionDiff(merged),
-            // ...extractRootServersDiffs(merged),
-            // ...extractChannelsDiffs(merged, operationChannel),
+            ...([...allOperationDiffs].filter(d => !otherMessageDiffs.has(d))),
```
b41ex: This will probable filter out more diffs than necessary in case of some changes in entity (e.g. payload) shared between different messages. We need to filter out diffs from other messages only if they are not related to the message for our APIHUB operation.

src/apitypes/async/async.changes.ts
```diff
@@ -137,18 +183,23 @@ export const compareDocuments: DocumentsCompare = async (
...
+            ...extractAsyncApiVersionDiff(merged),
```
b41ex: Changes to `id` and `defaultContentType` should also be explicitly included in the diffs.
Also please check that changes to the servers is collected in merged tree for the operation.

test/asyncapi-changes.test.ts
```diff
@@ -81,72 +62,267 @@ describe.skip('AsyncAPI 3.0 Changelog', () => {
       }, ASYNCAPI_API_TYPE))
     })
 
-    test('change operation', async () => {
-      const result = await buildChangelogPackageDefaultConfig('asyncapi-changes/operation/change')
+    test('should detect renamed operation as add and remove', async () => {
```
b41ex: This is the same as `add-remove` case, no need to duplicate.

test/asyncapi-changes.test.ts
```diff
@@ -81,72 +62,267 @@ describe.skip('AsyncAPI 3.0 Changelog', () => {
...
 
-      expect(result).toEqual(changesSummaryMatcher({ [ANNOTATION_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
-      expect(result).toEqual(numberOfImpactedOperationsMatcher({ [ANNOTATION_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
+      expect(result).toEqual(changesSummaryMatcher({ [BREAKING_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
+      expect(result).toEqual(numberOfImpactedOperationsMatcher({ [BREAKING_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
     })
 
-    test('renamed operation as add/remove', async () => {
-      const result = await buildChangelogPackageDefaultConfig('asyncapi-changes/operation/rename')
+    test('should not impact other operation when changing action type', async () => {
```
b41ex: The same as `change-action`, no need to duplicate.

test/asyncapi-changes.test.ts
```diff
@@ -81,72 +62,267 @@ describe.skip('AsyncAPI 3.0 Changelog', () => {
...
+        [ANNOTATION_CHANGE_TYPE]: 2,
       }, ASYNCAPI_API_TYPE))
     })
   })
 
-  describe('Servers', () => {
-    test('add server', async () => {
-      const result = await buildChangelogPackageDefaultConfig('asyncapi-changes/server/add')
+  describe('Channels tests', () => {
+    test('should detect changed operation channel reference', async () => {
```
b41ex: `should be tolerant to channel reference change`

test/asyncapi-changes.test.ts
```diff
@@ -81,72 +62,267 @@ describe.skip('AsyncAPI 3.0 Changelog', () => {
...
+      const result = await buildChangelogPackageDefaultConfig('asyncapi-changes/channel/change-reference')
+
+      expect(result).toEqual(changesSummaryMatcher({ [UNCLASSIFIED_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
+      expect(result).toEqual(numberOfImpactedOperationsMatcher({ [UNCLASSIFIED_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
+    })
+
+    test('should detect removed channel', async () => {
```
b41ex: Actually it is the same as previous case, no need to duplicate.

test/asyncapi-changes.test.ts
```diff
@@ -9,67 +9,48 @@ import {
...
-
-      expect(result).toEqual(changesSummaryMatcher({ [BREAKING_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
-      expect(result).toEqual(numberOfImpactedOperationsMatcher({ [BREAKING_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
-    })
+    expectNoChanges(result)
   })
 
   describe('Operations tests', () => {
-    test('add operation', async () => {
+    test('should detect added operation', async () => {
```
b41ex: I would use `report` term instead of `detect` everywhere in test names.
The logic is- `api-diff` is detecting changes, `api-processor` chooses to report changes or not.

test/asyncapi-changes.test.ts
```diff
@@ -81,72 +62,267 @@ describe.skip('AsyncAPI 3.0 Changelog', () => {
...
+    })
+
+    test('should detect changed channel address', async () => {
+      const result = await buildChangelogPackageDefaultConfig('asyncapi-changes/channel/change-address')
+
+      expect(result).toEqual(changesSummaryMatcher({ [UNCLASSIFIED_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
+      expect(result).toEqual(numberOfImpactedOperationsMatcher({ [UNCLASSIFIED_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
+    })
+
+    test('should detect added message definition in channel', async () => {
```
b41ex: Actually no changes should be reported for this case, since added message definition is not used in any of actual operations. Need to investigate why it is reported.

test/asyncapi-changes.test.ts
```diff
@@ -81,72 +62,267 @@ describe.skip('AsyncAPI 3.0 Changelog', () => {
...
+      const result = await buildChangelogPackageDefaultConfig('asyncapi-changes/channel/change-address-no-impact-on-other-channel')
+
+      // channel1 address changed, channel2 unchanged
+      // only operation1 (on channel1) should be impacted, not operation2 (on channel2)
+      expect(result).toEqual(changesSummaryMatcher({ [UNCLASSIFIED_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
+      expect(result).toEqual(numberOfImpactedOperationsMatcher({ [UNCLASSIFIED_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
     })
 
-    test('add root servers', async () => {
+    test('should detect added message definition in channel with multiple apihub operations', async () => {
```
b41ex: No changes should be reported in this case, since added message3 is not used in any actual operation.

test/asyncapi-changes.test.ts
```diff
@@ -81,72 +62,267 @@ describe.skip('AsyncAPI 3.0 Changelog', () => {
...
     })
 
-    test('remove root servers', async () => {
+    test('should not detect changes when removing root servers', async () => {
       const result = await buildChangelogPackageDefaultConfig('asyncapi-changes/server/remove-root')
 
+      expectNoChanges(result)
+    })
+
+    test('should not detect changes when changing root servers', async () => {
```
b41ex: Let's add similar test for modifying server in channel which is actually used in the operation.

test/asyncapi-changes.test.ts
```diff
@@ -81,72 +62,267 @@ describe.skip('AsyncAPI 3.0 Changelog', () => {
...
+      const result = await buildChangelogPackageDefaultConfig('asyncapi-changes/server/change-root')
+
+      expectNoChanges(result)
+    })
+  })
+
+  describe('Messages tests', () => {
+    test('should detect added message reference in operation', async () => {
```
b41ex: `should report added APIHUB operation when message reference is added to async operation`

test/asyncapi-changes.test.ts
```diff
@@ -81,72 +62,267 @@ describe.skip('AsyncAPI 3.0 Changelog', () => {
...
+      const result = await buildChangelogPackageDefaultConfig('asyncapi-changes/message/add-to-operation')
+
+      expect(result).toEqual(changesSummaryMatcher({ [BREAKING_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
+      expect(result).toEqual(numberOfImpactedOperationsMatcher({ [BREAKING_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
+    })
+
+    test('should detect removed message reference from operation', async () => {
```
b41ex: ditto

test/asyncapi-changes.test.ts
```diff
@@ -81,72 +62,267 @@ describe.skip('AsyncAPI 3.0 Changelog', () => {
...
+    test('should not impact other operation when adding message to one', async () => {
+      const result = await buildChangelogPackageDefaultConfig('asyncapi-changes/message/add-to-one-of-multiple-operations')
+
+      // message2 added to operation1, operation2 unchanged
+      // should only impact 1 new apihub operation (operation1-message2)
+      expect(result).toEqual(changesSummaryMatcher({ [BREAKING_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
+      expect(result).toEqual(numberOfImpactedOperationsMatcher({ [BREAKING_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
+    })
+
+    test('should not add changes to existing messages when adding new message to operation', async () => {
```
b41ex: The same as `add-to-one-of-multiple-operations`, no need to duplicate

test/asyncapi-changes.test.ts
```diff
@@ -81,72 +62,267 @@ describe.skip('AsyncAPI 3.0 Changelog', () => {
...
+    test('should only impact changed message when changing content type with multiple messages', async () => {
+      const result = await buildChangelogPackageDefaultConfig('asyncapi-changes/message/change-content-type-with-multiple-messages')
+
+      // Changing contentType of message1 in operation with message1 and message2
+      // should only impact operation1-message1, not operation1-message2
+      expect(result).toEqual(changesSummaryMatcher({ [BREAKING_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
+      expect(result).toEqual(numberOfImpactedOperationsMatcher({ [BREAKING_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
+    })
+
+    test('should not detect changes when removing unused message from channel', async () => {
```
b41ex: Same as `channel/remove-unused`, no need to duplicate.

test/asyncapi-changes.test.ts
```diff
@@ -81,72 +62,267 @@ describe.skip('AsyncAPI 3.0 Changelog', () => {
...
+      const result = await buildChangelogPackageDefaultConfig('asyncapi-changes/message/remove-unused-component-message')
+
+      // Removing a message from components/messages that is not referenced by any operation
+      // should not impact any apihub operations
+      expectNoChanges(result)
+    })
+  })
+
+  describe('Schema tests', () => {
+    test('should detect added property in message schema', async () => {
```
b41ex: Could be worth to have schema definitions itself in `components.schemas`, to check that changes to components are not reported,

b41ex (changes requested):
As discussed, let's add following group of tests:

BWC scopes

Change deduplication
- changes in shared entities
- different scopes
- in the same specification
- across different specifications

test/asyncapi-changes.test.ts
makeev-pavel: added components.schemas and ref

test/asyncapi-changes.test.ts
makeev-pavel: del

test/asyncapi-changes.test.ts
makeev-pavel: del

test/asyncapi-changes.test.ts
makeev-pavel: fixed

test/asyncapi-changes.test.ts
makeev-pavel: fixed

test/asyncapi-changes.test.ts
makeev-pavel: added 'should report changed server used in channel'

test/asyncapi-changes.test.ts
makeev-pavel: fixed

test/asyncapi-changes.test.ts
makeev-pavel: deleted

test/asyncapi-changes.test.ts
makeev-pavel: fixed

test/asyncapi-changes.test.ts
makeev-pavel: deleted

test/asyncapi-changes.test.ts
makeev-pavel: deleted

src/apitypes/async/async.changes.ts
makeev-pavel: added

makeev-pavel: and tests added

src/apitypes/async/async.changes.ts
makeev-pavel: fixed. added test 'should impact both operations when changing shared payload with multiple messages'

src/apitypes/async/async.changes.ts
makeev-pavel: fixed

test/asyncapi-changes.test.ts
makeev-pavel: fixed

test/asyncapi-changes.test.ts
makeev-pavel: fixed

test/asyncapi-changes.test.ts
```diff
@@ -81,72 +78,293 @@ describe.skip('AsyncAPI 3.0 Changelog', () => {
...
+      const result = await buildChangelogPackageDefaultConfig('asyncapi-changes/info/change-id')
 
       expect(result).toEqual(changesSummaryMatcher({ [ANNOTATION_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
       expect(result).toEqual(numberOfImpactedOperationsMatcher({ [ANNOTATION_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
     })
+
+    test('should report changed defaultContentType', async () => {
+      const result = await buildChangelogPackageDefaultConfig('asyncapi-changes/info/change-default-content-type')
+
+      // Root-level defaultContentType change + propagated breaking change in message contentType
```
b41ex: I think we should filter out changes to `defaultContentType` with scope root and only report its changes only if it really affects particular operation.

Also need to add test where there is defaultContentType is changed when there are two operations, one operation has explicit contentType (hence not affected by defaultContentTypeChange, no changes reported) and another without contentType (hence it is affected by defaultContentTypeChange) and it should be reported for it with corresponding scope.

src/apitypes/async/async.changes.ts
b41ex: For `id` it is OK, for defaultContentType the logic should be more complex, see new comment on 'should report changed defaultContentType' test.

test/asyncapi-changes.test.ts
b41ex: There's still `detect` in a couple of comments

test/asyncapi-changes.test.ts
```diff
@@ -81,72 +78,293 @@ describe.skip('AsyncAPI 3.0 Changelog', () => {
...
+      // operation1 and operation2 both reference channel1, address changed
+      // both apihub operations should be impacted
+      expect(result).toEqual(changesSummaryMatcher({ [UNCLASSIFIED_CHANGE_TYPE]: 2 }, ASYNCAPI_API_TYPE))
+      expect(result).toEqual(numberOfImpactedOperationsMatcher({ [UNCLASSIFIED_CHANGE_TYPE]: 2 }, ASYNCAPI_API_TYPE))
+    })
+
+    test('should not report changes when adding message definition in channel without referencing it in operation', async () => {
+      // message2 added to channel1.messages but operation1.messages still only references message1
+      // channel.messages add/remove diffs are filtered out — only operation.messages references matter
+      const result = await buildChangelogPackageDefaultConfig('asyncapi-changes/channel/add-message')
```
b41ex: `channel/add-message`, `channel/add-message-not-in-operation`, 'channel/add-message-with-multiple-operations' are essentially the same cases, no need to duplicate.

Let's have a test with samples currently in `channel/add-message`, but rename it to `channel/add-message-not-in-operation`, since this name is more fitting.

src/components/compare/bwc.validation.ts
```diff
@@ -158,3 +159,73 @@ export const createApihubApiCompatibilityScopeFunction = (
     return undefined
   }
 }
+
```
b41ex: I think it is better to place REST and Async parts in different files. And probably rename existing function to make more obvious that it is for REST.

src/apitypes/async/async.changes.ts
```diff
@@ -133,23 +151,31 @@ export const compareDocuments: DocumentsCompare = async (
...
+      }
+    }
+    return diffs
   }
 
   // Iterate through operations in merged document
-  const { operations } = merged
-  if (operations && isObject(operations)) {
-    for (const [asyncOperationId, operationData] of Object.entries(operations)) {
+  const { operations: asyncOperation } = merged
```
b41ex: Should be plural here, `asyncOperations`

src/components/compare/bwc.validation.ts
```diff
@@ -158,3 +159,73 @@ export const createApihubApiCompatibilityScopeFunction = (
...
+    const firstSegment = path?.[0]
+
+    // operations/<operationId>: resolve api-kind from operation x-api-kind with channel fallback
+    if (firstSegment === 'operations' && pathLength === ASYNC_OPERATION_PATH_LENGTH) {
+      // In normalized documents, operation.channel is the resolved channel object
+      const beforeChannelKind = getApiKindProperty((beforeJso as AsyncAPIV3.OperationObject | undefined)?.channel)
+      const afterChannelKind = getApiKindProperty((afterJso as AsyncAPIV3.OperationObject | undefined)?.channel)
+
+      // Operation's own x-api-kind takes priority, falls back to channel's x-api-kind
+      const beforeOperationKind = getApiKindProperty(beforeJso, beforeChannelKind) ?? APIHUB_API_COMPATIBILITY_KIND_BWC
```
b41ex: There should be no fallbacks here. As you remember, we designed this function to return `undefined` if parent scope should be kept. I thinks it should return `undefined` instead of fallback. See REST function, there's no fallbacks there.

src/components/compare/bwc.validation.ts
```diff
@@ -158,3 +159,73 @@ export const createApihubApiCompatibilityScopeFunction = (
...
+ * AsyncAPI has no document-level x-api-kind. The hierarchy is:
+ *   Channel (x-api-kind) → Operation (x-api-kind override) → Messages
+ *
+ * The scope function receives normalized before/after objects where $refs are resolved,
+ * so operation.channel is the resolved channel object with all its properties.
+ *
+ * - operations/<operationId>: operation x-api-kind overrides channel x-api-kind, defaults to bwc
+ * - channels/<channelId>: channel's own x-api-kind, defaults to bwc
+ */
+export const createAsyncApiCompatibilityScopeFunction = (): ApiCompatibilityScopeFunction => {
```
b41ex: I would make the same signature here as for the REST function- pass before/after documents api kind and used it to calculate `defaultApiCompatibilityKind`.
We might not have a means to specify api kind for AsyncAPI documents now, but I think it is worth having the same signature for future extensions.
Also, see how `defaultApiCompatibilityKind` is returned in REST function once for root scope and after that we don't need/should not have fallbacks in the code when calculating API kind.

src/components/compare/bwc.validation.ts
```diff
@@ -158,3 +159,73 @@ export const createApihubApiCompatibilityScopeFunction = (
...
+      if (beforeOperationKind === APIHUB_API_COMPATIBILITY_KIND_NO_BWC || afterOperationKind === APIHUB_API_COMPATIBILITY_KIND_NO_BWC) {
+        return API_COMPATIBILITY_KIND_NOT_BACKWARD_COMPATIBLE
+      }
+
+      return API_COMPATIBILITY_KIND_BACKWARD_COMPATIBLE
+    }
+
+    // channels/<channelId>: use channel's own x-api-kind
+    if (firstSegment === 'channels' && pathLength === ASYNC_CHANNEL_PATH_LENGTH) {
+      const beforeChannelKind = getApiKindProperty(beforeJso) ?? APIHUB_API_COMPATIBILITY_KIND_BWC
```
b41ex: Should be no fallbacks here

src/components/compare/bwc.validation.ts
```diff
@@ -158,3 +159,73 @@ export const createApihubApiCompatibilityScopeFunction = (
...
+      const afterChannelKind = getApiKindProperty(afterJso) ?? APIHUB_API_COMPATIBILITY_KIND_BWC
+
+      if (beforeChannelKind === APIHUB_API_COMPATIBILITY_KIND_NO_BWC || afterChannelKind === APIHUB_API_COMPATIBILITY_KIND_NO_BWC) {
+        return API_COMPATIBILITY_KIND_NOT_BACKWARD_COMPATIBLE
+      }
+
+      return API_COMPATIBILITY_KIND_BACKWARD_COMPATIBLE
```
b41ex: undefined

src/components/compare/bwc.validation.ts
```diff
@@ -158,3 +159,73 @@ export const createApihubApiCompatibilityScopeFunction = (
...
+    if (firstSegment === 'operations' && pathLength === ASYNC_OPERATION_PATH_LENGTH) {
+      // In normalized documents, operation.channel is the resolved channel object
+      const beforeChannelKind = getApiKindProperty((beforeJso as AsyncAPIV3.OperationObject | undefined)?.channel)
+      const afterChannelKind = getApiKindProperty((afterJso as AsyncAPIV3.OperationObject | undefined)?.channel)
+
+      // Operation's own x-api-kind takes priority, falls back to channel's x-api-kind
+      const beforeOperationKind = getApiKindProperty(beforeJso, beforeChannelKind) ?? APIHUB_API_COMPATIBILITY_KIND_BWC
+      const afterOperationKind = getApiKindProperty(afterJso, afterChannelKind) ?? APIHUB_API_COMPATIBILITY_KIND_BWC
+
+      const isRemoved = isObject(beforeJso) && !isObject(afterJso)
```
b41ex: Why the need to handle this case separately? In REST function it was due to PathItems/Paths logic, there's no such complication here.

test/asyncapi-apikind.test.ts
```diff
@@ -25,6 +30,137 @@ describe('AsyncAPI apiKind calculation', () => {
...
+          expect(scopeFunction([], {}, {})).toBe(API_COMPATIBILITY_KIND_BACKWARD_COMPATIBLE)
+        })
+
+        it('should return BWC for undefined path', () => {
+          expect(scopeFunction(undefined, {}, {})).toBe(API_COMPATIBILITY_KIND_BACKWARD_COMPATIBLE)
+        })
+      })
+
+      describe('Channels scope', () => {
+        it('should return BWC when channel has no x-api-kind', () => {
```
b41ex: Should return `undefined` for this case

test/asyncapi-build-apikind.test.ts
```diff
@@ -25,6 +30,137 @@ describe('AsyncAPI apiKind calculation', () => {
...
+        it('should return BWC for root path', () => {
+          expect(scopeFunction([], {}, {})).toBe(API_COMPATIBILITY_KIND_BACKWARD_COMPATIBLE)
+        })
+
+        it('should return BWC for undefined path', () => {
+          expect(scopeFunction(undefined, {}, {})).toBe(API_COMPATIBILITY_KIND_BACKWARD_COMPATIBLE)
+        })
+      })
+
+      describe('Channels scope', () => {
```
b41ex: 1. I thinks these kind of tests are better implemented by table-driven tests.
2. With the current selection of test data 2 combinations are missing (there are 3 possible values on 2 positions, so it should be 9 combinations, not 7).
3. I think we should also add to test data cases when before/after themselves are `undefined`, so here should be 16 cases to test.

test/asyncapi-apikind.test.ts
```diff
@@ -25,6 +30,137 @@ describe('AsyncAPI apiKind calculation', () => {
...
+
+        it('should return no-BWC when both channels have x-api-kind no-BWC', () => {
+          const before = { address: 'channel1', 'x-api-kind': 'no-BWC'}
+          const after = { address: 'channel1',  'x-api-kind': 'no-BWC'}
+          expect(scopeFunction(['channels', 'ch1'], before, after)).toBe(API_COMPATIBILITY_KIND_NOT_BACKWARD_COMPATIBLE)
+        })
+      })
+
+      describe('Operations scope', () => {
+        it('should return BWC when operation and channel have no x-api-kind', () => {
```
b41ex: ER: `undefined` for this case.

test/asyncapi-build-apikind.test.ts
```diff
@@ -25,6 +30,137 @@ describe('AsyncAPI apiKind calculation', () => {
...
+        })
+
+        it('should return no-BWC when both channels have x-api-kind no-BWC', () => {
+          const before = { address: 'channel1', 'x-api-kind': 'no-BWC'}
+          const after = { address: 'channel1',  'x-api-kind': 'no-BWC'}
+          expect(scopeFunction(['channels', 'ch1'], before, after)).toBe(API_COMPATIBILITY_KIND_NOT_BACKWARD_COMPATIBLE)
+        })
+      })
+
+      describe('Operations scope', () => {
```
b41ex: Ditto regarding table-driven tests.

test/asyncapi-build-apikind.test.ts
```diff
@@ -25,6 +30,137 @@ describe('AsyncAPI apiKind calculation', () => {
...
+        })
+
+        it('should return undefined for deeper channel paths', () => {
+          expect(scopeFunction(['channels', 'ch1', 'messages'], {}, {})).toBeUndefined()
+        })
+      })
+    })
   })
 
   describe('AsyncAPI operation/channel compatibility apiKind application', () => {
```
b41ex, line 33: 1. There are typos in `asyncapi/api-kind/base/spec.yaml`, 'cannel' -> 'channel'
2. It would be easier to read with naming like this (with kind specified explicitly)
 `operation<operationKind>Channel<channelKind>`, where kind is one of `NoKind`, `BWC`, `NoBWC`
NoBWC

test/asyncapi-changes.test.ts
```diff
@@ -153,64 +152,101 @@ describe('AsyncAPI 3.0 Changelog tests', () => {
...
+      expect(result).toEqual(numberOfImpactedOperationsMatcher({ [UNCLASSIFIED_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
+    })
+
+    test('should only impact operation whose channel references the changed server', async () => {
+      // Two channels, two servers: channel1→server1, channel2→server2.
+      // Only server1 host changed (api1 → new-api1).
+      // operation1 (on channel1) should get the diff, operation2 (on channel2) should not.
+      const result = await buildChangelogPackageDefaultConfig('asyncapi-changes/server/change-isolated-servers')
+
+      expect(result).toEqual(changesSummaryMatcher({ [UNCLASSIFIED_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
```
b41ex, line 252: For cases where we actually care about particular operations for which change is reported (like this one) it is better to use `operationChangesMatcher` and specify operationIds we are expecting.

test/asyncapi-deduplication.test.ts
```diff
@@ -0,0 +1,137 @@
...
+      // Same scope → diffs should be deduplicated by reference identity within one apiDiff call.
+      const result = await buildChangelogPackageDefaultConfig('asyncapi-deduplication/shared-schema-same-scope')
+
+      expect(result).toEqual(changesSummaryMatcher({ [BREAKING_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
+      expect(result).toEqual(numberOfImpactedOperationsMatcher({ [BREAKING_CHANGE_TYPE]: 2 }, ASYNCAPI_API_TYPE))
+    })
+  })
+
+  describe('Shared entities across different specifications', () => {
+    test('shared schema name in two specs, different scopes (receive vs send)', async () => {
```
b41ex: This single test is not enough, need to add:
- case when scopes are the same in different specifications, to actually test that duplicates are not counted
- case when schemas have the same name in different specifications, but different content. Schemas that do not have the same content should not be counted as the same schema and corresponding changes should not be deduplicated.

test/asyncapi-deduplication.test.ts
```diff
@@ -0,0 +1,137 @@
...
+      // operation1 (receive) in doc1, operation2 (send) in doc2.
+      // Both specs define SharedPayload with same change (number → string).
+      // Different apiDiff calls → different diff instances. Different operations → no cross-operation dedup.
```
b41ex: Comment is not correct. Deduplication based on instances won't work between different specs, but that means that deduplication is done using different means for this case.

test/asyncapi-deduplication.test.ts
```diff
@@ -0,0 +1,137 @@
...
+        [{ fileId: 'before1.yaml', publish: true }, { fileId: 'before2.yaml', publish: true }],
+        [{ fileId: 'after1.yaml' }, { fileId: 'after2.yaml' }],
+      )
+
+      expect(result).toEqual(changesSummaryMatcher({ [BREAKING_CHANGE_TYPE]: 2 }, ASYNCAPI_API_TYPE))
+      expect(result).toEqual(numberOfImpactedOperationsMatcher({ [BREAKING_CHANGE_TYPE]: 2 }, ASYNCAPI_API_TYPE))
+    })
+  })
+
+  describe('Root-level change deduplication', () => {
```
b41ex: I would propose to move this group under 'Shared entities in the same specification' group.

test/asyncapi-deduplication.test.ts
```diff
@@ -0,0 +1,137 @@
...
+      // No root-level diffs — server diffs come only through channel aggregation.
+      // unclassified: 2 in summary (one per scope), impacted 2.
+      const result = await buildChangelogPackageDefaultConfig('asyncapi-deduplication/root-server-change-multiple-operations')
+
+      expect(result).toEqual(changesSummaryMatcher({ [UNCLASSIFIED_CHANGE_TYPE]: 2 }, ASYNCAPI_API_TYPE))
+      expect(result).toEqual(numberOfImpactedOperationsMatcher({ [UNCLASSIFIED_CHANGE_TYPE]: 2 }, ASYNCAPI_API_TYPE))
+    })
+  })
+
+  describe('Message-level isolation', () => {
```
b41ex: 1. These tests are not about deduplication, the belong to asyncapi-changes file.
2. In cases like this it is better to check ER using `operationChangesMatcher` (not just summary numbers) to make sure we get changes in operations we expect.

test/asyncapi-deduplication.test.ts
```diff
@@ -0,0 +1,137 @@
...
+      }, ASYNCAPI_API_TYPE))
+      expect(result).toEqual(numberOfImpactedOperationsMatcher({
+        [ANNOTATION_CHANGE_TYPE]: 2,
+        [BREAKING_CHANGE_TYPE]: 1,
+      }, ASYNCAPI_API_TYPE))
+    })
+  })
+
+  describe('Cross-document deduplication', () => {
+    test('should deduplicate diffs when same operation appears in multiple document pairs', async () => {
```
b41ex: 1. I don't see a reason to have to groups of tests for very similar checks, just put the test into 'Shared entities across different specifications' group.
2. (separate topic)Sample case has different specifications which has operations with the same operationId. This should actually trigger our duplicated operation check and result in warning. Btw, for the AsyncAPI I would say let us make this not warning, but error. We couldn't do it for REST, because we already have specifications published with more tolerant rules, but we could still do it for AsyncAPI.

b41ex (changes requested):

src/components/compare/bwc.validation.ts
```diff
@@ -158,3 +159,73 @@ export const createApihubApiCompatibilityScopeFunction = (
...
 }
+
+const ASYNC_OPERATION_PATH_LENGTH = 2 // operations/<operationId>
+const ASYNC_CHANNEL_PATH_LENGTH = 2   // channels/<channelId>
+
+/**
+ * Creates an ApiCompatibilityScopeFunction for AsyncAPI documents.
+ *
+ * AsyncAPI has no document-level x-api-kind. The hierarchy is:
+ *   Channel (x-api-kind) → Operation (x-api-kind override) → Messages
```
JayLim2: is it by design? because right is `operation -> channel -> messages`

test/asyncapi-changes.test.ts
```diff
@@ -81,72 +78,293 @@ describe.skip('AsyncAPI 3.0 Changelog', () => {
...
+
+  describe('Messages tests', () => {
+    test('should report added APIHUB operation when message reference is added to async operation', async () => {
+      const result = await buildChangelogPackageDefaultConfig('asyncapi-changes/message/add-to-operation')
+
+      expect(result).toEqual(changesSummaryMatcher({ [UNCLASSIFIED_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
+      expect(result).toEqual(numberOfImpactedOperationsMatcher({ [UNCLASSIFIED_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
+    })
+
+    test('should report removed APIHUB operation when message reference is removed to async operation', async () => {
```
JayLim2: `removed from`

test/projects/asyncapi-changes/channel/add-message-with-multiple-operations/after.yaml
```diff
@@ -0,0 +1,42 @@
...
+    address: channel1
+    messages:
+      message1:
+        $ref: '#/components/messages/message1'
+      message2:
+        $ref: '#/components/messages/message2'
+      message3:
+        $ref: '#/components/messages/message3'
+operations:
+  operation1:
```
JayLim2, line 15: I see only 1 operation. Why `add-message-with-**multiple**-operations`?

test/projects/asyncapi-changes/server/change-in-channel/after.yaml
```diff
@@ -0,0 +1,31 @@
...
+info:
+  title: Test AsyncAPI
+  version: 1.0.0
+servers:
+  production:
+    protocol: amqp
+    host: api.production.example.com
+channels:
+  channel1:
+    address: channel1
```
JayLim2, line 11: Please, highlight for me, what's change? address, servers, messages didn't change
even message instance is not changed

src/components/compare/bwc.validation.ts
makeev-pavel: By design
```
Resolution order:
AsyncAPI structure is Channel → Operation → Message: a channel contains operations, and each operation may reference messages.
```
Operation level can override channel apiKind

test/projects/asyncapi-changes/server/change-in-channel/after.yaml
makeev-pavel: The change is in servers.production.host (line 8):                                                                                                                                                                                                                                                                
  - before: api.example.com                                                                                                                                                                                                                                                                                         
  - after: api.production.example.com                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                                                    
`channel1 `references this server via $ref: '#/servers/production', so the server host change propagates to the channel and its operation through the reference.

test/projects/asyncapi-changes/channel/add-message-with-multiple-operations/after.yaml
makeev-pavel: deleted

test/asyncapi-changes.test.ts
makeev-pavel: fixed

test/asyncapi-changes.test.ts
makeev-pavel: done

src/components/compare/bwc.validation.ts
makeev-pavel: Created
`bwc.validation.async.ts`
`bwc.validation.rest.ts`
`bwc.validation.types.ts`

src/apitypes/async/async.changes.ts
makeev-pavel: done

src/components/compare/bwc.validation.ts
makeev-pavel: fixed

src/components/compare/bwc.validation.ts
makeev-pavel: fixed

src/components/compare/bwc.validation.ts
makeev-pavel: fixed

src/components/compare/bwc.validation.ts
makeev-pavel: fixed

src/components/compare/bwc.validation.ts
makeev-pavel: deleted

test/asyncapi-apikind.test.ts
makeev-pavel: fixed

test/asyncapi-build-apikind.test.ts
makeev-pavel: I fixed. But we need to check if that's it

test/asyncapi-apikind.test.ts
makeev-pavel: The matrix has been added

test/asyncapi-build-apikind.test.ts
makeev-pavel: done

test/asyncapi-build-apikind.test.ts
makeev-pavel: fixed, fixed

test/asyncapi-changes.test.ts
makeev-pavel: added

test/asyncapi-deduplication.test.ts
makeev-pavel: added new cases

test/asyncapi-deduplication.test.ts
makeev-pavel: done

test/asyncapi-deduplication.test.ts
makeev-pavel: changed to
```
operation1 (receive) in doc1, operation2 (send) in doc2.
Both specs define SharedPayload with same change (number → string).
Different scopes → separate apiDiff calls → separate diff instances.
Cross-document content-based dedup (calculateDiffId) applies per operation,
but these are different operations so both changes are counted.
```

test/asyncapi-deduplication.test.ts
makeev-pavel: moved and used operationChangesMatcher

test/asyncapi-deduplication.test.ts
makeev-pavel: moved

Added new `DuplicateOperationHandler`

src/components/compare/bwc.validation.ts
b41ex: I would say we usually follow pattern where api type is in the beginning of the file name.
Let's rename first two to:
- `async.bwc.validation.ts` and
- `rest.bwc.validation.ts`

test/asyncapi-changes.test.ts
makeev-pavel: Added new logic. You need to validate
comment
https://github.com/Netcracker/qubership-apihub-api-processor/pull/70/changes/67c31b0167ac2dfa4935aade40106402792d7324
Added test
'should not report defaultContentType change for message with explicit contentType'

test/asyncapi-apikind.test.ts
```diff
@@ -25,68 +31,218 @@ describe('AsyncAPI apiKind calculation', () => {
...
+          )
+          expect(scopeFunction([], {}, {})).toBe(expected)
+        })
+      })
+
+      describe('Channels scope', () => {
+        it.each([
+          // default | before | after | expected
+          ['bwc', 'undefined', 'undefined', undefined],
+          ['bwc', 'undefined', 'bwc', undefined],
```
b41ex: `undefined` as ER for cases where one of the value is `undefined` and another is `bwc` (or even worse- both are `bwc`, but the ER is `undefined`) looks counter-intuitive to me. I would say that in these cases the function should return `bwc`.

test/asyncapi-apikind.test.ts
```diff
@@ -25,68 +31,218 @@ describe('AsyncAPI apiKind calculation', () => {
...
+          const scopeFunction = createAsyncApiCompatibilityScopeFunction(documentApiKind)
+          expect(scopeFunction(['channels', 'ch1'], createChannel(beforeKind), createChannel(afterKind))).toBe(expected)
+        })
+      })
+
+      describe('Operations scope', () => {
+        it.each([
+          // default | before | after | expected
+          ['bwc', 'undefined', 'undefined', undefined],
+          ['bwc', 'undefined', 'bwc', undefined],
```
b41ex: Ditto regarding ER `undefined` for cases 
- `undefined`, `bwc`
- `bwc`, `undefined`
- `bwc`, `bwc`

test/asyncapi-build-apikind.test.ts
b41ex: Do not see cases when `beforeJso`/`afterJso` are `undefined`.

test/asyncapi-changes.test.ts
b41ex: I would say most of the tests in this file should use these matcher.
We do not care about particular changes for particular operations only when checking summary e.g. in deduplication tests, but these are in different file.

test/asyncapi-deduplication.test.ts
```diff
@@ -0,0 +1,124 @@
...
+    })
+  })
+
+  describe('Shared entities across different specifications', () => {
+    test('shared schema name in two specs, different scopes (receive vs send)', async () => {
+      // operation1 (receive) in doc1, operation2 (send) in doc2.
+      // Both specs define SharedPayload with same change (number → string).
+      // Different scopes → separate apiDiff calls → separate diff instances.
+      // Cross-document content-based dedup (calculateDiffId) applies per operation,
+      // but these are different operations so both changes are counted.
```
b41ex: 'but these operations have different scope, so both changes are counted'

test/asyncapi-deduplication.test.ts
```diff
@@ -0,0 +1,124 @@
...
+        'asyncapi-deduplication/shared-schema-cross-specs-different-content',
+        [{ fileId: 'before1.yaml', publish: true }, { fileId: 'before2.yaml', publish: true }],
+        [{ fileId: 'after1.yaml' }, { fileId: 'after2.yaml' }],
+      )
+
+      expect(result).toEqual(changesSummaryMatcher({ [BREAKING_CHANGE_TYPE]: 2 }, ASYNCAPI_API_TYPE))
+      expect(result).toEqual(numberOfImpactedOperationsMatcher({ [BREAKING_CHANGE_TYPE]: 2 }, ASYNCAPI_API_TYPE))
+    })
+
+    test('should throw error when same operationId appears in multiple documents', async () => {
```
b41ex: This test is not about deduplication, I would put it into `asyncapi-changes.test`
Btw, the error should be triggered during `build` type publication, need to have this test as well.

src/apitypes/async/async.changes.ts
```diff
@@ -211,19 +159,24 @@ export const compareDocuments: DocumentsCompare = async (
 
           const otherMessageDiffs = collectExclusiveOtherMessageDiffs(messages, messageIndex)
           const channelMessageDiffs = collectChannelMessageDefinitionDiffs(operationChannel as AsyncAPIV3.ChannelObject)
+          // defaultContentType only affects messages without explicit contentType
+          const messageAffectedByDefaultContentType =
+            !hasExplicitContentType(prevDocData, messageId) ||
```
b41ex: This condition is incorrect. We only need to check if `currDocData` has explicit content type. If it does- then effective content type is determined by it, so, defaultContentType changes are not relevant. They are only relevant if `currDocData` does not have explicit content type.
Need to make sure there are tests for both cases.

src/apitypes/async/async.utils.ts
```diff
@@ -271,6 +271,76 @@ export const resolveAsyncApiOperationIdsFromRefs = (
   return resolved
 }
 
+/**
+ * Checks whether a message has explicit `contentType` in the raw (pre-normalization) document.
+ * Messages without explicit contentType inherit from `defaultContentType`.
+ */
+export function hasExplicitContentType(doc: AsyncAPIV3.AsyncAPIObject | undefined | null, messageId: string): boolean {
+  const message = getKeyValue(doc, 'components', 'messages', messageId)
```
b41ex: This is incorrect logic. The message might not be defined in the `components` section at all or it could be defined via reference chain etc.

Need to check specific message in the normalized document.

b41ex (changes requested):

src/components/compare/bwc.validation.ts
makeev-pavel: done

test/asyncapi-changes.test.ts
makeev-pavel: done

test/asyncapi-changes.test.ts
makeev-pavel: fixed

test/asyncapi-apikind.test.ts
makeev-pavel: discussed separately
https://github.com/Netcracker/qubership-apihub-api-processor/pull/72

test/asyncapi-apikind.test.ts
makeev-pavel: discussed separately
https://github.com/Netcracker/qubership-apihub-api-processor/pull/72

test/asyncapi-deduplication.test.ts
makeev-pavel: fixed

test/asyncapi-deduplication.test.ts
makeev-pavel: added

src/apitypes/async/async.changes.ts
makeev-pavel: I delete messageAffectedByDefaultContentType. You need to check final logic
Commit https://github.com/Netcracker/qubership-apihub-api-processor/pull/70/changes/757832d0579b135198c5ea9018b4b5c34b9e2eb0

src/apitypes/async/async.utils.ts
makeev-pavel: I delete hasExplicitContentType. You need to check final logic
Commit https://github.com/Netcracker/qubership-apihub-api-processor/commit/757832d0579b135198c5ea9018b4b5c34b9e2eb0

test/asyncapi-deduplication.test.ts
```diff
@@ -0,0 +1,115 @@
...
+ * Within one document pair (by reference identity):
+ *   - `aggregateDiffsWithRollup` propagates diffs bottom-up via Set<Diff>
+ *   - Same Diff instance can appear in multiple operations (e.g. shared component schema)
+ *   - `comparePairedDocs` deduplicates via `new Set(allDiffs)` — reference identity
+ *
+ * Across multiple document pairs (by content hash):
+ *   - Rare case: one operation in multiple documents → multiple apiDiff() calls
+ *   - Uses `removeObjectDuplicates(diffs, calculateDiffId)` for content-based dedup
+ */
+describe('AsyncAPI deduplication tests', () => {
```
b41ex, line 22: It is worth adding deduplication test for `defaultContentType` changes- several messages without explicit content type set, defaultContentType changes, check that only one diff is reported in summary.

test/asyncapi-changes.test.ts
```diff
@@ -1,152 +1,582 @@
...
+      expect(result).toEqual(changesSummaryMatcher({ [BREAKING_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
+      expect(result).toEqual(operationChangesMatcher([
+        expect.objectContaining({
+          operationId: 'operation2-message2',
+          previousOperationId: 'operation2-message2',
+        }),
+      ]))
+    })
+
+    test('should only affect message without explicit contentType when defaultContentType changes in same operation', async () => {
```
b41ex, line 530: The case is the same as in change-default-content-type-with-explicit-override`. No need to duplicate, remove one of the tests.

b41ex (changes requested):

JayLim2 (approved):

test/asyncapi-build-apikind.test.ts
```diff
@@ -25,68 +32,228 @@ describe('AsyncAPI apiKind calculation', () => {
         expect(result).toBe(expected)
       })
     })
+
+    describe('Changelog backward compatibility scope function', () => {
```
b41ex: Integration tests are missing for the apiKind handling during `changelog` build.
Integration tests are important here, because final result depends on behavior of both compatibility scope function provided by api-processor and behavior of api-diff.
Review corresponding integration tests for REST at least in these test groups and add analogous tests for Async.
'ApiKind operations section tests'
'Priority operation ApiKind and default ApiKind'
'Remove operations tests'
There is actually less places where we could set api kind in Async compared to REST, so number of tests for Async will be less than number of tests for REST, but cases we need to check are all similar.

b41ex (changes requested):

test/asyncapi-build-apikind.test.ts
makeev-pavel: Added integration tests for AsyncAPI x-api-kind handling in asyncapi-changelog-apikind.test.ts, analogous to the REST tests in apiKinds.test.ts.                                                                                                                                
                                                                                                                                                                                                                                                                                  
**Operation-level x-api-kind (9 cases, table-driven via test.each):**                                                                                                                                                                                                               
  All transitions between none, BWC, and no-BWC in prev/curr documents — verifies that no-BWC in either document results in risky classification, and BWC (or default) results in breaking.

**Channel-level x-api-kind (9 cases, same table reused):**
  Same transition matrix applied at channel level.

**Operation + Channel priority (4 cases):**
  - Fallback: operation without x-api-kind inherits from channel (channel-bwc-operation-none, channel-nobwc-operation-none)
  - Override: operation x-api-kind takes priority over channel (channel-bwc-operation-nobwc, channel-nobwc-operation-bwc)

**Remove operation (7 cases):**
  - Removed operation with x-api-kind: BWC → breaking
  - Removed operation with x-api-kind: no-BWC → risky
  - Removed operation without x-api-kind → breaking (default BWC)
  - Priority: operation x-api-kind overrides channel on removal (remove-channel-nobwc-operation-bwc, remove-channel-bwc-operation-nobwc)
  - Fallback: removed operation without x-api-kind uses channel (remove-channel-nobwc-operation-none, remove-channel-bwc-operation-none)

_Note on "add operation": not covered because adding an operation is classified as non-breaking by api-diff regardless of x-api-kind — same as in REST tests (apiKinds.test.ts)._

test/asyncapi-changes.test.ts
makeev-pavel: del

test/asyncapi-deduplication.test.ts
makeev-pavel: added

test/asyncapi-changelog-apikind.test.ts
```diff
@@ -0,0 +1,108 @@
...
+      expect(result).toEqual(numberOfImpactedOperationsMatcher({ [RISKY_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
+    })
+
+    test('should prioritize operation no-BWC over channel BWC', async () => {
+      const result = await buildChangelogPackageDefaultConfig('asyncapi-changelog-apikind/operation/channel-bwc-operation-nobwc')
+      expect(result).toEqual(changesSummaryMatcher({ [RISKY_CHANGE_TYPE]: 1, [UNCLASSIFIED_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
+      expect(result).toEqual(numberOfImpactedOperationsMatcher({ [RISKY_CHANGE_TYPE]: 1, [UNCLASSIFIED_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
+    })
+
+    test('should prioritize operation BWC over channel no-BWC', async () => {
```
b41ex: This is inconsistent with what we have for REST. 
See corresponding test REST test: `should prioritize parent no-BWC in previous document over operation BWC property in current document`

test/asyncapi-changelog-apikind.test.ts
```diff
@@ -0,0 +1,108 @@
...
+    ['should apply BWC when x-api-kind: BWC removed in current document',                          'bwc-to-none',    { [BREAKING_CHANGE_TYPE]: 1, [UNCLASSIFIED_CHANGE_TYPE]: 1 }],
+    ['should apply BWC when x-api-kind: BWC in both documents',                                    'bwc-to-bwc',     { [BREAKING_CHANGE_TYPE]: 1 }],
+    ['should apply no-BWC when x-api-kind changed from BWC to no-BWC',                             'bwc-to-nobwc',   { [RISKY_CHANGE_TYPE]: 1, [UNCLASSIFIED_CHANGE_TYPE]: 1 }],
+    ['should apply no-BWC when x-api-kind: no-BWC in previous document and removed in current',    'nobwc-to-none',  { [RISKY_CHANGE_TYPE]: 1, [UNCLASSIFIED_CHANGE_TYPE]: 1 }],
+    ['should apply no-BWC when x-api-kind changed from no-BWC to BWC',                             'nobwc-to-bwc',   { [RISKY_CHANGE_TYPE]: 1, [UNCLASSIFIED_CHANGE_TYPE]: 1 }],
+    ['should apply no-BWC when x-api-kind: no-BWC in both documents',                              'nobwc-to-nobwc', { [RISKY_CHANGE_TYPE]: 1 }],
+  ]
+
+  describe('Operation-level x-api-kind', () => {
```
b41ex: These cases are trivial and essentially tested by unit tests for api compatibility function.
I don't see much value to have integration tests for these cases.

test/asyncapi-changelog-apikind.test.ts
```diff
@@ -0,0 +1,108 @@
...
+    runApiKindCases('operation', apiKindTransitionCases)
+  })
+
+  describe('Channel-level x-api-kind', () => {
```
b41ex: These cases are trivial and essentially tested by unit tests for api compatibility function.
I don't see much value to have integration tests for these cases.

test/asyncapi-changelog-apikind.test.ts
```diff
@@ -0,0 +1,108 @@
...
+    runApiKindCases('channel', apiKindTransitionCases)
+  })
+
+  describe('Operation + Channel x-api-kind', () => {
```
b41ex: There are a total of 81 possible cases for before-channel/before-operation/after-channel/after-operation api kind options (undefined, noBwc, bwc).
But most of them are trivial (e.g. both before/after operations has explicit api kind of both have undefined api kind).
Really interesting cases (and specific for changelog, they are not covered by unit tests for api compatibility function) are kind of criss-crossed cases, when explicit channel api kind is replaced with explicit operation api kind and vise versa.
| before-channel-apiKind | before-operation-apiKind |after-channel-apiKind| after-operation-apiKind |ER |
| --- | --- | --- | --- | --- |
|undefined|noBwc|bwc|undefined|noBwc|
|undefined|bwc|noBwc|undefined|noBwc|
|noBwc|undefined|undefined|bwc|noBwc|
|bwc|undefined|undefined|noBwc|noBwc|

I would say these are the tests worth testing via integrations tests.
It is curious that only one of these cases are currently tested (`should prioritize operation BWC over channel no-BWC`) and it has incorrect ER (inconsistent with REST), see comment below.

test/asyncapi-changelog-apikind.test.ts
```diff
@@ -0,0 +1,108 @@
...
+    })
+
+    test('should prioritize operation BWC over channel no-BWC', async () => {
+      const result = await buildChangelogPackageDefaultConfig('asyncapi-changelog-apikind/operation/channel-nobwc-operation-bwc')
+      expect(result).toEqual(changesSummaryMatcher({ [BREAKING_CHANGE_TYPE]: 1, [UNCLASSIFIED_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
+      expect(result).toEqual(numberOfImpactedOperationsMatcher({ [BREAKING_CHANGE_TYPE]: 1, [UNCLASSIFIED_CHANGE_TYPE]: 1 }, ASYNCAPI_API_TYPE))
+    })
+  })
+
+  describe('Remove operation tests', () => {
```
b41ex, line 246: 1. Since we have a folder `remove` there seems to be no need to have `remove` in cases names
2. It would really simplify reviewing and understanding the code if case name would include all relevant information (before channel api-kind, before operation api kind, after channel api kind), like `before-channel-undefined-operation-bwc-after-channel-undefined`
3. We need to add cases when whole channel is removed (this is analogous to `Remove pathItem tests` for REST.
4. I do not get cases selection logic, why current cases are tested and these are not?

|	before-channel	|	before-operation	|	after-channel	|
|	---	|	---	|	--- |
|	undefined	|	undefined	|	noBwc	|
|	undefined	|	undefined	|	bwc	|
|	undefined	|	noBwc	|	noBwc	|
|	undefined	|	noBwc	|	bwc	|
|	undefined	|	bwc	|	noBwc	|
|	undefined	|	bwc	|	bwc	|
|	noBwc	|	undefined	|	undefined	|
|	noBwc	|	undefined	|	bwc	|
|	noBwc	|	noBwc	|	undefined	|
|	noBwc	|	noBwc	|	noBwc	|
|	noBwc	|	noBwc	|	bwc	|
|	noBwc	|	bwc	|	undefined	|
|	noBwc	|	bwc	|	bwc	|
|	bwc	|	undefined	|	undefined	|
|	bwc	|	undefined	|	noBwc	|
|	bwc	|	noBwc	|	undefined	|
|	bwc	|	noBwc	|	noBwc	|
|	bwc	|	bwc	|	undefined	|
|	bwc	|	bwc	|	noBwc	|
|	bwc	|	bwc	|	bwc	|

b41ex (changes requested):

test/asyncapi-changelog-apikind.test.ts
makeev-pavel: by design:
`For API Kind, channel-level x-api-kind is the default for all APIHUB operations on that channel; operation-level x-api-kind overrides when present (more specific overrides more general, same as in REST)`

then it is necessary to make corrections to the design that No-BWC cannot be redefined.

test/asyncapi-changelog-apikind.test.ts
makeev-pavel: These are logic tests that show what the result will be if there is a certain apiKind

test/asyncapi-changelog-apikind.test.ts
makeev-pavel: above, as already written

test/asyncapi-changelog-apikind.test.ts
makeev-pavel: it makes no sense to test all 81 combinations. It is checked separately in the tests
- channel calculation
- operations calculation
- the final result of channel + operation

test/asyncapi-changelog-apikind.test.ts
makeev-pavel: 1. Deleted
2. I don't see the point in testing before-channel + after-channel + before-operation + after-operation 
3. Yes, added
4. As I wrote above, the calculation of a separate parent, child and their interaction (priority)  is being tested separately. For rest, we also don't test all possible options.

makeev-pavel: After discussion, it was decided to add a complete table of all possible combinations. Added

test/asyncapi-changelog-apikind.test.ts
makeev-pavel: After discussion, it was decided to add a complete table of all possible combinations. Added

b41ex (approved):

PR #76, opened 2026-04-08
test: use in-memory fs to store build results during tests

makeev-pavel:
Improves test reliability and enables easier debugging of generated files
- introduce registryFs abstraction to unify fs access in tests
- default to in-memory filesystem for performance and isolation
- add optional disk mode via FS_MODE=disk for debugging
- implement utility helpers for reading files and configs from registryFs
- add test:disk script using Jest setup file to enable disk mode
- normalize path handling between memfs and real fs


test/document-group.test.ts
```diff
@@ -288,7 +288,7 @@ describe('Document Group test', () => {
         const pkg = LocalRegistry.openPackage(`document-group/${folder}/not-hang-up-when-processing-cycled-chain-for-response`, groupToOnePathOperationIdsMap)
         await pkg.publish(pkg.packageId, { packageId: pkg.packageId })
 
-        const notificationFile = await loadFileAsString(
+        const notificationFile = await vfsLoadFileAsString(
```
b41ex: I don't like vfs leaking on this level of abstraction.
And also depending on flag with which tests were started it could either be real fs or memfs (vfs).
I would propose a different split, like `loadFileAsStringFromRegistry` for cases when we need to read file from build result?

test/helpers/registry/fs.ts
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * Build a path through vfs — absolute on disk, posix in memory.
+ */
+export function vfsPath(...segments: string[]): string {
+  return useDisk
+    ? path.resolve(process.cwd(), ...segments)
+    : path.posix.join(...segments)
+}
+
+export { vfs }
```
b41ex: I would say that name `vfs` leaks abstraction here and it is also misleading, since it could be both real fs and memfs.
I would call it without exposing implementation details, e.g. `registryFs`, `testFs` or even just `fs` (although the latter is probable also easy to confuse with standard fs)

test/document-group.test.ts
b41ex: Let's also add jsdoc on both `loadFileAsString` and `loadFileAsStringFromRegistry` (an any other similar pair of functions) explaining when to use one or another.

b41ex (changes requested):

test/helpers/registry/fs.ts
makeev-pavel: done

test/document-group.test.ts
makeev-pavel: done

b41ex (approved):

PR #77, opened 2026-04-08
fix: filter channel messages per operation in createOperationSpec for AsyncAPI

### LINKED ISSUE #569: [Bug][FE] deprecated operator for schema objects doesn't consider its value (true or false)
### Describe the bug

Preconditions: AsyncAPI spec with both options for deprecated operator  (true and false)

[depr in payload.yaml](https://github.com/user-attachments/files/26370561/depr.in.payload.yaml)

### To Reproduce

1. Publish spec from preconditions
2. Navigate to the Deprecated tab

**ER: only one schema object is defined as deprecated (where deprecated: true)
AR: both schema objects are defined as deprecated (deprecated: true\false)**

### Version

Frontend v. 3.3.1-dev.20260318091141
API Processor v. 5.0.1-dev.20260318090240

### Logs

_No response_

### Additional information

Image

makeev-pavel:
- Fixed deprecated items leaking between AsyncAPI operations that share the same channel but use different messages                                                                                                                                                                                              
- Extracted `filterChannelMessages` utility in `async.utils.ts` to filter channel messages per operation in `createOperationSpec`                                                                                                                                                                                  
- Added tests for deprecated items isolation and channel message filtering

### DIFF
--- a/src/apitypes/async/async.operation.ts
+++ b/src/apitypes/async/async.operation.ts
@@ -50,6 +50,7 @@ import {
   createBaseAsyncApiSpec,
   enrichAsyncApiWithInlineRefs,
   extractProtocol,
+  filterChannelMessages,
   getAsyncMessageId,
   isMessageObject,
   resolveAsyncApiOperationIdsFromRefs,
@@ -275,7 +276,11 @@ export const createOperationSpec = (
       )
 
       if (requestedIdsSet.has(calculatedId)) {
-        selectedOperations[asyncOperationId] = operationObject
+        selectedOperations[asyncOperationId] = {
+          ...operationObject,
+          messages: [message],
+          channel: filterChannelMessages(channel as AsyncAPIV3.ChannelObject, messageId),
+        }
         break
       }
     }
--- a/src/apitypes/async/async.operation.ts
+++ b/src/apitypes/async/async.operation.ts
@@ -292,7 +292,19 @@ export const createOperationSpec = (
     }
   }
 
-  return createBaseAsyncApiSpec(normalizedDocument, selectedOperations)
+  // Build root channels record from filtered channels, keyed by their original name
+  let channels: Record<string, AsyncAPIV3.ChannelObject> | undefined
+  if (filteredChannels.size > 0 && normalizedDocument.channels) {
+    channels = {}
+    for (const [channelName, channelObj] of Object.entries(normalizedDocument.channels)) {
+      const filteredChannel = filteredChannels.get(channelObj as AsyncAPIV3.ChannelObject)
+      if (filteredChannel) {
+        channels[channelName] = filteredChannel
+      }
+    }
+  }
+
+  return createBaseAsyncApiSpec(normalizedDocument, selectedOperations, channels)
 }
 
 /**
--- a/src/apitypes/async/async.operation.ts
+++ b/src/apitypes/async/async.operation.ts
@@ -44,13 +44,14 @@ import {
 } from '@netcracker/qubership-apihub-api-unifier'
 import { calculateHash, ObjectHashCache } from '../../utils/hashes'
 import {
+  addMessageToOperation,
   buildAsyncApiSpecFromDocument,
   calculateAsyncApiKind,
   checkHasAsyncApiOperations,
   createBaseAsyncApiSpec,
+  createOperationWithSingleMessage,
   enrichAsyncApiWithInlineRefs,
   extractProtocol,
-  filterChannelMessages,
   getAsyncMessageId,
   isMessageObject,
   resolveAsyncApiOperationIdsFromRefs,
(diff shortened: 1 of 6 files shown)


src/apitypes/async/async.utils.ts
```diff
@@ -131,6 +131,17 @@ export const getAsyncChannelId = (channel: AsyncAPIV3.ChannelObject): string =>
   return getAsyncObjectId(channel)
 }
 
+export const filterChannelMessages = (
+  channel: AsyncAPIV3.ChannelObject,
+  messageId: string,
+): AsyncAPIV3.ChannelObject => {
+  const channelMessages = (channel as AsyncAPIV3.ChannelObject).messages
```
iurii-golovinskii: Maybe it works without casting?

src/apitypes/async/async.operation.ts
```diff
@@ -275,7 +276,11 @@ export const createOperationSpec = (
       )
 
       if (requestedIdsSet.has(calculatedId)) {
-        selectedOperations[asyncOperationId] = operationObject
+        selectedOperations[asyncOperationId] = {
+          ...operationObject,
+          messages: [message],
+          channel: filterChannelMessages(channel as AsyncAPIV3.ChannelObject, messageId),
```
b41ex: In my understanding this code does not work correctly when `createOperationSpec` is called with more than one operationId. In this case we expect to get specification which contains parts relevant for all corresponding operations, but here we filter for one messageId only.
Also there's no corresponding tests. Need to have a test where specification has at least 3 operations/messages with shared channel and we call `createOperationSpec` with 2 of them.

src/apitypes/async/async.utils.ts
```diff
@@ -131,6 +131,17 @@ export const getAsyncChannelId = (channel: AsyncAPIV3.ChannelObject): string =>
...
+  if (!isObject(channelMessages)) {
+    return channel
+  }
+  return { ...channel, messages: { [messageId]: (channelMessages as AsyncAPIV3.MessagesObject)[messageId] } }
```
b41ex: Same, casting is excessive.

test/asyncapi-deprecated.test.ts
```diff
@@ -79,6 +79,26 @@ describe('AsyncAPI 3.0 Deprecated tests', () => {
...
+      expect(operation1Entry).toBeDefined()
+      const operation1DeprecatedItems = operation1Entry![1].deprecatedItems ?? []
+
+      // operation2 uses UserQuit (deprecated: true on email) — should have deprecated items
+      const operation2Entry = operations.find(([key]) => key.includes('operation2'))
+      expect(operation2Entry).toBeDefined()
+      const operation2DeprecatedItems = operation2Entry![1].deprecatedItems ?? []
+
+      expect(operation1DeprecatedItems.length).toBe(0)
+      expect(operation2DeprecatedItems.length).toBeGreaterThan(0)
```
b41ex: Let's check the exact number just in case.

b41ex (changes requested):

src/apitypes/async/async.utils.ts
makeev-pavel: rewrote the code, removed the casting

src/apitypes/async/async.utils.ts
makeev-pavel: rewrote the code, removed the casting

test/asyncapi-deprecated.test.ts
makeev-pavel: done

src/apitypes/async/async.operation.ts
makeev-pavel: fixed + add test

test/asyncapi-operation.test.ts
```diff
@@ -358,6 +358,135 @@ describe('AsyncAPI 3.0 Operation Tests', () => {
...
+        operationIdC = calculateAsyncOperationId(MULTI_MESSAGE_OP, MESSAGE_ID_C)
+        operationIdD = calculateAsyncOperationId(OPERATION_D, MESSAGE_ID_D)
+        operationIdE = calculateAsyncOperationId(OPERATION_E, MESSAGE_ID_E)
+        operationIdF = calculateAsyncOperationId(OPERATION_F, MESSAGE_ID_A)
+
+        const sharedChannelDoc = await loadYamlFile<AsyncAPIV3.AsyncAPIObject>('asyncapi/operations/shared-channel/spec.yaml')
+        sharedChannelNormalized = normalizeAsyncApiDocument(sharedChannelDoc)
+      })
+
+      test('should include only the relevant message in channel when requesting single operationId', () => {
```
b41ex, line 409: As discussed, resulting specification should include root `channels` property with relevant channels and relevant messages under this channels.
ER for tests should be extended to also check this root channels.

b41ex (changes requested):

test/asyncapi-operation.test.ts
makeev-pavel: done

b41ex (approved):

PR #81, opened 2026-04-20
feat: add Model Context Protocol contracts support

### LINKED ISSUE #576: [FE] MCP API type support - api-processor

makeev-pavel:
## Summary

  Adds first-class support for **MCP (Model Context Protocol)** artifacts to the
  api-processor build pipeline. MCP servers publish non-API artifacts — `init`,
  `tools`, `resources`, and `prompts` — which are now built, indexed, and packaged
  alongside regular API operations within the same versioned package model.

  MCP is implemented as a new API type (`mcp`), parallel to REST/GraphQL/AsyncAPI,
  reusing the existing `ApiBuilder` abstraction so the rest of the pipeline stays
  kind-agnostic.

  ## What's new

  ### New API type & document types
  - New `mcpBuilder` (`src/apitypes/mcp/`) registered as a regular `ApiBuilder`.
  - Four document types detected **structurally from JSON shape** (no manual
    tagging by the caller): `mcp-init`, `mcp-tools`, `mcp-resources`, `mcp-prompts`
    (`detectMcpDocumentType` in `mcp.parser.ts`).
  - Each source file is a single MCP type; the endpoint is supplied per file via
    `metadata.mcpEndpoint`.

  ### Archive / output format
  - Top-level **`mcp.json`** index, grouped by kind: `{ inits, tools, resources, prompts }`.
  - Per-entity payload files under **`mcp/{mcpEntityId}`** (no extension, matching
    the operations data-file idiom).
  - `groupMcpEntitiesByKind` strips `data` from the index so payloads don't leak
    into `mcp.json`.

  ### Validation
  - Hard failure when an MCP file is missing `metadata.mcpEndpoint`.
  - Duplicate detection within a file and across files sharing an endpoint.
  - **Capability cross-check**: warns (non-fatal) when `init` declares a capability
    (`tools`/`resources`/`prompts`) but no matching entities are found on that
    endpoint (`validateMcpCapabilities`).
  - List items without a `name` are skipped and surfaced as notifications instead
    of producing degenerate IDs.
  - JSON schemas for the four doc types are included (`src/apitypes/mcp/schemas/`);
    schema violations are **reported, not fatal**, and the entity is still built.

  ### Cross-version compatibility
  - Field reads are stable across MCP protocol revisions (2024-11-05 … 2025-11-25).
  - `title` (added 2025-06-18) is preferred for display, falling back to `name`;
    IDs are always derived from `name`, so they stay stable across versions.

  ## Testing
  - `test/mcp-unit.test.ts` — parser, ID calculation, capability cross-check, and
    payload-wrapping unit tests (inline fixtures).
  - `test/mcp-build.test.ts` — end-to-end build tests: document-type detection,
    per-kind entity extraction, payload shape, ID format, duplicate handling,
    multi-endpoint publishing, incremental add/remove/change, mixed MCP + REST
    builds, and the full publish → `mcp.json` index + `mcp/{id}` payload round-trip.
  - Registry test helpers extended to persist MCP entities (`saveMcpEntities`).
  - **67 tests, all passing.**

  ## Breaking changes
  - `BuildResult` / `BuildResultDto` expose MCP entities via `buildResult.mcpEntities`
    (single map; entities carry `.data`). Consumers reading MCP output should use this.


src/apitypes/mcp/mcp.parser.ts
```diff
@@ -0,0 +1,161 @@
...
+  }))
+}
+
+function detectMcpDocumentType(obj: Record<string, unknown>): string | undefined {
+  if (isObject(obj.capabilities) && isString(obj.protocolVersion) && isObject(obj.serverInfo)) {
+    return MCP_DOCUMENT_TYPE.MCP_INIT
+  }
+  if (Array.isArray(obj.tools)) { return MCP_DOCUMENT_TYPE.MCP_TOOLS }
+  if (Array.isArray(obj.resources)) { return MCP_DOCUMENT_TYPE.MCP_RESOURCES }
+  if (Array.isArray(obj.prompts)) { return MCP_DOCUMENT_TYPE.MCP_PROMPTS }
```
JayLim2, line 45: 1. Is it impossible when several of fields `tools`, `resources`, `prompts` are present simultaneously?
2. shouldn't we check that they aren't empty?

src/apitypes/mcp/mcp.parser.ts
```diff
@@ -0,0 +1,161 @@
...
+
+type McpListItemRaw = Record<string, unknown> & McpNamedItem
+
+function hasValidName(item: Record<string, unknown>): item is McpListItemRaw {
+  return isString(item.name) && item.name.trim().length > 0
+}
+
+interface ExtractEntitiesResult {
+  entities: McpEntityRaw[]
+  errors: { message: string }[]
```
JayLim2: type alias maybe?

src/apitypes/mcp/mcp.parser.ts
```diff
@@ -0,0 +1,161 @@
...
+  if (docType === MCP_DOCUMENT_TYPE.MCP_INIT) {
+    // init is always exactly one entity, but we deliberately wrap it in an array so that
+    // ParsedMcpData.entities stays a uniform McpEntityRaw[] across all doc types. This keeps
+    // the rest of the pipeline (buildMcpEntities, id/title calc, dedup, grouping) kind-agnostic —
+    // a single `.map` over entities, with no init-vs-list branching downstream.
+    return { entities: [{ kind: MCP_KIND.INIT, name: 'initialize', data: obj }], errors: [] }
+  }
+
+  const mapping = MCP_LIST_TYPE_MAPPING[docType]
+  if (!mapping) { return { entities: [], errors: [] } }
```
JayLim2: can't we extract this to constant?

src/apitypes/mcp/mcp.parser.ts
```diff
@@ -0,0 +1,161 @@
...
+
+  const arr = obj[mapping.key]
+  if (!Array.isArray(arr)) { return { entities: [], errors: [] } }
+
+  const entities: McpEntityRaw[] = []
+  const errors: { message: string }[] = []
```
JayLim2: type alias

JayLim2: same that above

src/apitypes/mcp/mcp.parser.ts
```diff
@@ -0,0 +1,161 @@
...
+    parsed = JSON.parse(data)
+  } catch {
+    // MCP document type is detected from the parsed object's shape, so at this point
+    // we don't yet know whether this is an MCP file. Invalid JSON simply means it cannot
+    // be one — return undefined to let the next builder in the chain claim the file
+    // (parseFile rethrows on a thrown error, which would abort the whole parse chain).
+    return undefined
+  }
+
+  if (!isObject(parsed) || Array.isArray(parsed)) {
```
JayLim2: check it, i think `isObject` also checks that it is not an array

src/apitypes/mcp/mcp.parser.ts
```diff
@@ -0,0 +1,161 @@
...
+    return undefined
+  }
+
+  const originalDocument = parsed as Record<string, unknown>
```
JayLim2: and if so, you can get rid of type assertion

src/apitypes/unknown/index.ts
```diff
@@ -23,7 +23,7 @@ import { DOCUMENT_TYPE } from '../../consts'
 export * from './unknown.consts'
 export { buildBinaryDocument } from './unknown.document'
 
-export const unknownApiBuilder: ApiBuilder<string> = {
+export const unknownApiBuilder: ApiBuilder<unknown> = {
```
JayLim2, line 26: ???

src/apitypes/unknown/unknown.document.ts
```diff
@@ -85,7 +88,7 @@ export const buildBinaryDocument: (parsedFile: SourceFile, file: BuildConfigFile
   }
 }
 
-export const dumpUnknownDocument: DocumentDumper<string> = (document) => {
+export const dumpUnknownDocument: DocumentDumper<unknown> = (document) => {
```
JayLim2, line 91: ???

src/apitypes/unknown/index.ts
makeev-pavel: It's a correctness fix that surfaced while aligning all api types onto the same typed ApiBuilder interface.
The unknown api type was declared as ApiBuilder<string>, meaning parsedFile.data was typed as string. That was wrong: for JSON/YAML files the parser yields a parsed object, not a string. The old code parsedFile.data?.description only "worked" because the incorrect string type let property access slide through without a real guarantee.
When the builder generics were tightened to register MCP alongside REST/GraphQL/AsyncAPI in one heterogeneous ApiBuilder[], the bogus <string> no longer fit. Changing it to <unknown> reflects reality

src/apitypes/mcp/mcp.parser.ts
makeev-pavel: 1. Multiple of tools/resources/prompts at once: by design each MCP source file is a single document type. This is part of the publish contract — the publisher provides separate files for init, tools, resources, and prompts, each tagged via metadata.mcpEndpoint. So a file carrying several of these arrays is not an expected/valid input. Detection is deliberately first-match-wins (tools → resources → prompts) and we treat the file as that one type. 
2. Detection only classifies the type; whether the document actually produces anything is handled one step later. extractEntities returns zero entities for an empty array, and parseMcpFile then returns undefined when entities.length === 0, so the file is not claimed as MCP. There's a unit test for exactly this ('empty tools array' → undefined). So an empty tools: [] doesn't slip through as a valid MCP document.

src/apitypes/mcp/mcp.parser.ts
makeev-pavel: yes. added

src/apitypes/mcp/mcp.parser.ts
makeev-pavel: added

src/apitypes/mcp/mcp.parser.ts
makeev-pavel: added

src/apitypes/mcp/mcp.parser.ts
makeev-pavel: fixed

src/apitypes/mcp/mcp.parser.ts
makeev-pavel: deleted

src/apitypes/unknown/unknown.document.ts
makeev-pavel: answered above

test/mcp-build.test.ts
```diff
@@ -0,0 +1,622 @@
...
+      expect(entitiesOfKind(result, 'tool')[0].title).toBe('valid_tool')
+
+      // the skipped item is surfaced as a notification rather than silently swallowed
+      const note = result.notifications.find(n => /missing or empty required 'name'/.test(n.message))
+      expect(note).toBeDefined()
+      expect(note!.fileId).toBe('tools-partial-invalid.json')
+    })
+  })
+
+  describe('Incremental update', () => {
```
b41ex: Incremental update (`update` method) is not used anymore in production use cases, it was used by Editor which was removed. We should plan to remove the `update` method. For now- let us not add new functionality there and then fix potential bugs.

test/mcp-build.test.ts
```diff
@@ -0,0 +1,622 @@
...
+        ],
+      })
+      expectEntityCounts(initial, { init: 1, tool: 2, resource: 2 })
+
+      const updated = await editor.update(
+        { files: [{ fileId: 'resources.json', metadata: { mcpEndpoint: MCP_ENDPOINT } }] },
+        [],
+      )
+
+      expectEntityCounts(updated, { resource: 2 })
```
b41ex: E.g. here we check (incorrect) behavior that when using `update` method we can remove `init` artefact from MCP, thus producing build result which bypasses mandatory validation.

src/components/mcp.ts
```diff
@@ -0,0 +1,132 @@
...
+export function groupMcpEntitiesByKind(entities: McpEntityIndex): PackageMcpFile {
+  const grouped: PackageMcpFile = { inits: [], tools: [], resources: [], prompts: [] }
+  for (const { data: _data, ...index } of entities.values()) {
+    // strip the payload (`data`) — mcp.json is the lightweight index; payloads live in mcp/{id}.json
+    grouped[KIND_TO_FIELD[index.kind]].push(index)
+  }
+  return grouped
+}
+
+export function validateMcpInitRequired(entities: McpEntityIndex): void {
```
b41ex, line 101: validateMcpInitRequired only checks whether any `init` exists anywhere, so `/mcp/a` can satisfy tools under `/mcp/b`.
Need to check that each MCP endpoint has `init` artefact.

src/apitypes/mcp/mcp.parser.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  const originalDocument: Record<string, unknown> = parsed
+  const docType = detectMcpDocumentType(originalDocument)
+  if (!docType) {
+    return undefined
+  }
+
+  const { entities, errors } = extractEntities(originalDocument, docType)
+  if (entities.length === 0) {
+    return undefined
```
b41ex: MCP files with only invalid list items now falls through as unknown.
I think a detected MCP file should still be added as MCP file and report extraction/schema errors, even when all list items are invalid.

src/apitypes/mcp/mcp.entities.ts
```diff
@@ -0,0 +1,84 @@
...
+
+// the init entity has no per-item title; it is shown as a fixed "Overview" label (kept as "init" in data)
+const INIT_ENTITY_TITLE = 'init'
+
+export function calculateMcpEntityId(
+  mcpEndpoint: string,
+  kind: string,
+  name: string,
+): string {
+  const safeEndpoint = slugify(mcpEndpoint, SLUG_OPTIONS_OPERATION_ID).replace(/^-+|-+$/g, '')
```
b41ex: Why this `.replace(/^-+|-+$/g, '')`? We don't have it for other entity IDs and it reduces selectivity.

src/apitypes/mcp/schemas/mcp-init.json
```diff
@@ -0,0 +1,31 @@
+{
```
b41ex: Need to use official schemas verbatim, for each supported version.

test/helpers/registry/local.ts
```diff
@@ -357,14 +359,18 @@ export class LocalRegistry implements IRegistry {
     return undefined
   }
 
-  private getDocApiTypeGuard(apiType: OperationsApiType): (document: ZippableDocument | ResolvedVersionDocument) => void {
+  private getDocApiTypeGuard(apiType: OperationsApiType | string): (document: ZippableDocument | ResolvedVersionDocument) => void {
```
b41ex: I would add `ContractsType` instead of raw string.

test/mcp-unit.test.ts
```diff
@@ -0,0 +1,293 @@
...
+    expect(result!.errors!.some(e => /expected an object/.test(e.message))).toBe(true)
+  })
+
+  describe('list extraction', () => {
+    // build a minimal valid item of the given kind (each list type has its own required fields)
+    const makeItem = (kind: McpKind, name: string): Record<string, unknown> => {
+      switch (kind) {
+        case 'resource': return { uri: `file:///${name}`, name }
+        case 'tool': return { name, inputSchema: { type: 'object' } }
+        default: return { name } // prompt
```
b41ex: I would have explicit case for `prompt`, would be better if we support other entity types in the future.

test/mcp-build.test.ts
```diff
@@ -0,0 +1,622 @@
...
+      expect(withTitle?.title).toBe('Nice Title')
+
+      // no title → title falls back to the name; no description → empty string
+      const noDesc = entitiesOfKind(result, 'tool').find(t => t.mcpEntityId === 'mcp-tool-no_desc')
+      expect(noDesc?.title).toBe('no_desc')
+      expect(noDesc?.description).toBe('')
+    })
+  })
+
+  describe('Entity ID format', () => {
```
b41ex: I think unit tests for entityId calculation are enough, no need for e2e tests

test/mcp-build.test.ts
```diff
@@ -0,0 +1,622 @@
...
+        files: [
+          initFile(),
+          { fileId: 'tools.json', metadata: { mcpEndpoint: MCP_ENDPOINT } },
+        ],
+      })
+
+      const tool = entitiesOfKind(result, 'tool').find(t => t.title === 'search_docs')
+      expect(tool?.data).toEqual({
+        tools: [{
+          name: 'search_docs',
```
b41ex: This object is checked at least twice, may be extract to constant?

test/mcp-build.test.ts
```diff
@@ -0,0 +1,622 @@
...
+      await expect(
+        editor.run({
+          files: [
+            { fileId: 'tools.json' },
+          ],
+        }),
+      ).rejects.toThrow(/mcpEndpoint/)
+    })
+
+    test('should report a schema violation (tool missing inputSchema) but still build the entity', async () => {
```
b41ex: I see this in the design, why not error, but notification?

Using JSON schema for MCP spec
- support all released versions
- **invalid input breaks publish**

b41ex (changes requested):

test/mcp-build.test.ts
makeev-pavel: Agreed — removed the MCP incremental-update tests

test/mcp-build.test.ts
makeev-pavel: removed. Mandatory MCP validation (per-endpoint init + schema) runs on the build path

src/components/mcp.ts
makeev-pavel: Fixed — validateMcpInitRequired is now per-endpoint: every endpoint that publishes entities must carry its own init, otherwise the publish fails. Added a multi-endpoint test.

src/apitypes/mcp/mcp.parser.ts
makeev-pavel: Fixed

src/apitypes/mcp/mcp.entities.ts
makeev-pavel: fixed

src/apitypes/mcp/schemas/mcp-init.json
makeev-pavel: After discussion, we came to the conclusion that we are not doing this now

test/helpers/registry/local.ts
makeev-pavel: fixed

test/mcp-unit.test.ts
makeev-pavel: yes, fixed

test/mcp-build.test.ts
makeev-pavel: removed the e2e entityId tests; the calculation is covered by the unit tests (MCP entity ID table)

test/mcp-build.test.ts
makeev-pavel: Done — extracted to SEARCH_DOCS_TOOL_PAYLOAD

test/mcp-build.test.ts
makeev-pavel: Fixed — schema non-conformance now breaks the publish

PR #82, opened 2026-04-24
fix: filter channels to selected operations in create operation spec with refs

### LINKED ISSUE #626: [Bug][FE]: Incorrect channel list when exporting Async API
### Describe the bug

[base.yaml](https://github.com/user-attachments/files/27152877/base.yaml)

### To Reproduce

1. Publish AsyncApi spec 
**Required**: 2+ operations and 2+ channels 
**Example**:
```
operations:
  operation1:
    channel:
      $ref: '#/channels/channel1'
      ...
  operation2:
    channel:
      $ref: '#/channels/channel2'
    ...
```
2. Create Operation Groups with 1 operation
3. Call export

**ER**: Export spec contain only operation 1 and channel 1
**AR**: Export spec contain only operation 1, channel 1, **channel 2**

### Version

Latest

### Logs

_No response_

### Additional information

_No response_

makeev-pavel:


src/apitypes/async/async.operation.ts
```diff
@@ -241,7 +240,7 @@ const collectDeprecatedItems = (
  * - a message in source `operations` has no resolvable id
  */
 export const createOperationSpec = (
-  normalizedDocument: AsyncAPIV3.AsyncAPIObject,
+  normalizedDocument: AsyncAPIV3.AsyncAPIObject | TYPE.AsyncOperationData,
```
JayLim2, line 243: Is this new type from 3rd-party?

JayLim2 (approved):

src/apitypes/async/async.operation.ts
makeev-pavel: AsyncAPIV3 - "@asyncapi/parser"
TYPE.AsyncOperationData - internal type form async.types.ts

JayLim2 (approved):

b41ex (changes requested):
Required: `createOperationSpecEnrichedWithRefs` now passes a cropped `refsOnlySingleOperationSpec` into `buildAsyncApiSpecFromDocument`, but `buildAsyncApiSpecFromDocument` still assumes `refsDocument.operations[op].messages` is positionally aligned with the original source operation messages.

```typescript
  const resultSpec = buildAsyncApiSpecFromDocument(document, matchedMessageRefsByOperation, refsOnlySingleOperationSpec)
  enrichAsyncApiDocumentWithRefs(resultSpec, document, refsOnlySingleOperationSpec)
```
```typescript
    const sourceMessages = sourceOperation.messages || []
    const refsMessages = (refsOperations[asyncOperationId]?.messages as AsyncAPIV3.MessageObject[] | undefined) ?? []
    const filteredMessages: (AsyncAPIV3.MessageObject | AsyncAPIV3.ReferenceObject)[] = []
    sourceMessages.forEach((message, index) => {
      if (isReferenceObject(message) && matchedRefs.has(message.$ref)) {
        filteredMessages.push(message)
        const resolvedMessage = refsMessages[index]
```

For an operation with messages `[A, B]`, selecting only `B` makes `createOperationSpec(refsDocument, opIdB)` produce `refsMessages === [B]`. The source loop still visits `B` at `index === 1`, so `refsMessages[1]` is `undefined`. That breaks `resolvedMessages` collection and can incorrectly omit root `defaultContentType` for non-first selected messages. The PR’s new tests only cover `OPERATION_ID_1`, so they miss this case.

makeev-pavel:
**Fix:** `createOperationSpec `produces a compact `refsMessages `array containing only selected messages (e.g., selecting `B` from `[A, B, C]` yields `[B]` at `index 0`). The source loop iterates all messages, so `B` sits at positional `index 1` — causing a mismatch. The separate counter stays aligned with the compact array because matched messages appear in the same relative order in both arrays. Added a test that selects a non-first message (`MessageB`) from a multi-message operation and verifies `defaultContentType `propagation — it fails without the fix.

b41ex (approved):

PR #84, opened 2026-04-30
chore: update Node.js to v.24

### LINKED ISSUE #619: [FE] Update NodeJS to latest LTS version (24.x)
Since NodeJS 20 is at the EOL , we need to update NodeJS version to the latest LTS (24.x) and update other dependencies accordingly (e.g. npm to v. 11.x etc.)
Start with the most potentially problematic repositories (apispec-view, ui) to identify potential problems but the full scope is all APIHUB FE components.
For building some components (e.g. apispec-view )we use  https://github.com/netcracker/qubership-apihub-nodejs-dev-image . It should also be updated with the latest LTS versions.

makeev-pavel:
### Node.js 24 Migration                                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                                                    
  **Summary**                                 
                                                                                                                                                                                                                                                                                                                    
  - Upgrade Node.js runtime from 20 to 24                                                                                                                                                                                                                                                                           
  - Switch CI workflow (frontend-ci) to feature/nodejs24 branch of qubership-apihub-ci
  - Switch internal dependencies (api-diff, api-unifier, graphapi, json-crawl, npm-gitflow) to feature-nodejs24 dist-tag                                                                                                                                                                                            
  - Update @types/node from ^18 to ^24                         
  - Update engines.node from >=18.0.0 to >=24.0.0
  - Bump TypeScript target from es2021 to es2022 — required because the codebase uses Array.prototype.at() (an ES2022 feature) in async.utils.ts and rest.bwc.validation.ts; with updated dependencies TypeScript now correctly enforces the target library check
  (https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping)

  **Test plan**

  - CI frontend-ci workflow passes (build + unit tests)

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -4,7 +4,7 @@
     "resolveJsonModule": true,
     /* Basic Options */
     // "incremental": true,                   /* Enable incremental compilation */
-    "target": "es2021",
+    "target": "es2022",
     /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */
     "module": "commonjs",
     /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', 'es2020', or 'ESNext'. */
--- a/.github/workflows/automatic-pr-labeler.yaml
+++ b/.github/workflows/automatic-pr-labeler.yaml
@@ -10,9 +10,9 @@ name: Automatic PR Labeler
 
 on:
   pull_request:
-    branches: [develop]
+    branches: [ develop ]
     types:
-      [opened, reopened, synchronize]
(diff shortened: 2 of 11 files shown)


tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "resolveJsonModule": true,
     /* Basic Options */
     // "incremental": true,                   /* Enable incremental compilation */
-    "target": "es2021",
+    "target": "es2022",
```
makeev-pavel, line 7: With the dependency update for Node 24, TypeScript now correctly reports https://github.com/microsoft/TypeScript/wiki/Node-Target-Mapping: Property 'at' does not exist on type 'string[]'. Do you need to change your target library? Try changing the 'lib' compiler
  option to 'es2022' or later.

  Bumping target to es2022 is safe — Node.js 20+ (and 24) fully supports ES2022.

b41ex (approved):

PR #87, opened 2026-05-15
feat: ability to start changelog build with different levels of api-processor version validation

### LINKED ISSUE #635: [Feat]: Enhance builder version check to support migration mode
### Is your feature request related to a problem? Please describe

Current builder version check blocks publication process until migration is not completed.
It directly impacts Apihub operations and not acceptable.

### Describe the solution you'd like

Need to modify existing builder version check. New version should consider migration running. 

If migration is in progress, then publication should be allowed even if previous version is built with old builder (not migrated yet). If the This case need to add a special version metadata field called `previousBuilderVersion`. 
But this rule is applicable only in case of minor or patch builder version change.
Examples:
* 1.0.0 -> 1.0.1 - continue build
* 1.0.0 -> 1.1.0 - continue build
* 1.0.0 -> 2.0.0 - fail build

Migration status should be retrieved from `/api/v1/system/info`, property `migrationInProgress`.

# Scope/design

## api-processor scope
Definitions: current api-processor version === version of the api-processor currently building the changelog

Add `apiProcessorVersionValidationLevel` (optional) to BuilderRunOptions, possible values are:
- `major`
- `strict` (default)

Note: this should be added to BuilderRunOptions, not to BuildConfig, since it is a so it is a transient option for the current build try, not something we want persisted in the BuildConfig.

Modify the `validateApiProcessorVersion` so that 
- if `apiProcessorVersionValidationLevel` is `major`, validation fails only if major part of the builderVersion for previous or current package version differs from the major part of current api-processor version
  - if any of versions are not parseable as [SemVer](https://semver.org/), validation should not fail
- if `apiProcessorVersionValidationLevel` is `strict`, validation fails if builderVersion for previous or current package version differs from current api-processor version (current behaviour)

Store build version for package versions participating in the comparison if they differ from the current api-processor version
- if `builderVersion` for the previous package version differs from the current api-processor version, store `builderVersion` for the previous package version to the `previousVersionBuilderVersion` in the `info.json` of the changelog build result. Do not add `previousVersionBuilderVersion` if it the same as current api-processor version.
-if `builderVersion` for the current package version differs from the current api-processor version, store `builderVersion` for the current package version to the `currentVersionBuilderVersion` in the `info.json` of the changelog build result. Do not add `currentVersionBuilderVersion` if it the same as current api-processor version.

So, overall info.json:
- will have `builderVersion` holding current api-processor version
- may have `previousVersionBuilderVersion`
- may have `currentVersionBuilderVersion`

Document new properties in the API specification

## UI scope
If migration is in progress (as reported by `/api/v1/system/info`, property `migrationInProgress`), start changelog builds with `apiProcessorVersionValidationLevel` set to `major`

## Build task consumer 
Same as UI

makeev-pavel:
## Summary

  - Added `apiProcessorVersionValidationLevel` option to `BuilderRunOptions` to support relaxed version validation during data migration
  - In `major` mode, changelog builds are allowed when the major version matches, even if minor/patch differs — this unblocks users from publishing while migration is in progress                                                                                                           
  - When versions participating in a comparison were built by a different api-processor, their builder versions are recorded in `previousVersionBuilderVersion` / `currentVersionBuilderVersion` fields of the build result, so the backend can identify and rebuild them after migration completes

  ## Changes

  ### Types
  - **`src/types/internal/builder.ts`**: Added `VersionValidationLevel` type (`'major' | 'strict'`) and `apiProcessorVersionValidationLevel` to `BuilderRunOptions`
  - **`src/types/internal/compare.ts`**: Added `apiProcessorVersionValidationLevel` to `CompareContext`, new `BuilderVersionInfo` and `CompareResult` types
  - **`src/types/external/config.ts`**, **`src/types/package/config.ts`**: Added `previousVersionBuilderVersion` and `currentVersionBuilderVersion` fields to `BuildConfig` and `PackageConfig`

  ### Validation logic (`src/validators.ts`)
  - **`validateApiProcessorVersion`**: Added `validationLevel` parameter (default `strict`). In `major` mode: compares only major version part, skips validation if either version is unparseable
  - **`getMismatchedBuilderVersion`**: New helper — returns the builder version from resolved version data only if it differs from the current api-processor version
  - **`applyBuilderVersionInfo`**: New helper — copies `previousVersionBuilderVersion` / `currentVersionBuilderVersion` from comparison result to build config (for persistence in `info.json`)

  ### Comparison pipeline
  - **`src/components/compare/compare.operations.ts`**: `compareVersionsOperations` now returns `BuilderVersionInfo` alongside the comparison. Passes `validationLevel` from context to `validateApiProcessorVersion`. Uses `getMismatchedBuilderVersion` to detect version mismatches
  - **`src/components/compare/compare.ts`**: `compareVersions` returns `CompareResult` (comparisons + builder version info). Builder version info from ref-packages is discarded (only relevant at root level)

  ### Strategy integration
  - **`src/builder.ts`**, **`src/strategies/build.strategy.ts`**, **`src/strategies/changelog.strategy.ts`**, **`src/strategies/prefix-groups-changelog.strategy.ts`**: All call sites updated to destructure `CompareResult` and apply builder version info to config via `applyBuilderVersionInfo`

  ### Test infrastructure
  - **`test/helpers/registry/local.ts`**: Added `VersionOverrideRegistry` — extends `LocalRegistry` to override `apiProcessorVersion` per version in `versionResolver`
  - **`test/helpers/utils.ts`**: Added `buildChangelogWithVersionOverrides` helper for e2e tests with configurable validation level and spec content

  ### Tests (`test/validators.test.ts`)
  - **28 unit tests** for `validateApiProcessorVersion`: parametrized matrix covering strict/major modes, exact matches, minor/patch/major mismatches, prerelease suffixes, unparseable versions, null versions, error prefix formatting
  - **6 e2e tests** through `PackageVersionBuilder`: strict pass/fail, major pass/fail for both previous and current version sides
  - **4 integration tests** for builder version info: verifies `previousVersionBuilderVersion` and `currentVersionBuilderVersion` are set/unset correctly in changelog build results


src/validators.ts
```diff
@@ -30,14 +30,52 @@ export function validateConfig(config: BuildConfig): void {
   }
 }
 
-export function validateApiProcessorVersion(resolvedVersion: VersionCache | null, errorPrefix?: string): void {
+export function validateApiProcessorVersion(
+  resolvedVersion: VersionCache | null,
+  errorPrefix?: string,
+  validationLevel: VersionValidationLevel = 'strict',
```
b41ex: I would propose to not spread defaults in several places, imo it is better to have it defined in one place, and you already have it applied in `compare.operations.ts`

src/validators.ts
```diff
@@ -30,14 +30,52 @@ export function validateConfig(config: BuildConfig): void {
...
+}
+
+export function getMismatchedBuilderVersion(versionData: VersionCache | null | undefined): string | undefined {
+  if (!versionData?.apiProcessorVersion || versionData.apiProcessorVersion === apiProcessorVersion) {
+    return undefined
   }
+  return versionData.apiProcessorVersion
+}
+
+export function applyBuilderVersionInfo(config: PackageConfig, source: BuilderVersionInfo): void {
```
b41ex, line 69: There is a has a stale metadata bug here.
Because the helper only writes truthy values, it never clears `previousVersionBuilderVersion` or `currentVersionBuilderVersion` when a later comparison no longer has a mismatch. 
This can leave stale “built by different api-processor” markers after affected versions are rebuilt or after a rebuild compares matching processor versions. Let's clear both fields first, then set the fields present in CompareResult, and add a regression test for stale field removal.

b41ex (changes requested):

src/types/internal/builder.ts
```diff
@@ -71,6 +71,8 @@ export interface IPackageVersionBuilder {
 
 export type FileSourceMap = Record<string, Blob>
 
+export type VersionValidationLevel = 'major' | 'strict'
```
JayLim2: enum or constants, use constants instead of literals everywhere needed

src/types/package/config.ts
```diff
@@ -28,4 +28,7 @@ export interface PackageConfig {
 
   metadata?: Record<string, unknown>
   format?: ExportFormat
+
+  previousVersionBuilderVersion?: string
+  currentVersionBuilderVersion?: string
```
JayLim2, line 33: isn't this type declared?

src/types/external/config.ts
```diff
@@ -74,6 +74,9 @@ export interface BuildConfig extends BuildConfigBase {
   // todo since it goes straight to the info.json, remove it from every buildconfig except the one that is 'build'
   validationRulesSeverity?: ValidationRulesSeverity
   operationsSpecTransformation?: OperationsSpecTransformation
+
+  previousVersionBuilderVersion?: string
+  currentVersionBuilderVersion?: string
```
JayLim2, line 79: isn't this type declared?

src/types/internal/compare.ts
```diff
@@ -126,4 +127,14 @@ export interface CompareContext {
   versionDocumentsResolver: VersionDocumentsResolver
   rawDocumentResolver: _RawDocumentResolver
   normalizedSpecFragmentsHashCache: ObjectHashCache
+  apiProcessorVersionValidationLevel?: VersionValidationLevel
 }
+
+export type BuilderVersionInfo = {
+  previousVersionBuilderVersion?: string
+  currentVersionBuilderVersion?: string
+}
```
JayLim2, line 136: here it is declared)

JayLim2 (changes requested):

src/types/package/config.ts
makeev-pavel: this is standard practice here. There will be a lot of unnecessary  code to create inheritance or switch from interfaces to types. I would leave it as it is.

BuildConfig and PackageConfig have the same fields

src/types/internal/builder.ts
makeev-pavel: done

src/validators.ts
makeev-pavel: done

src/validators.ts
makeev-pavel: done

JayLim2 (approved):

b41ex (approved):

PR #101, opened 2026-07-17
fix: omit tags not used by any REST operation when exporting operation group

### LINKED ISSUE #721: [Bug][BE] Empty Tags in OpenAPI export output files
### Describe the bug

Empty Tags in OpenAPI export output

### To Reproduce

when user exports API group, and group does not contain operations connected with specific tag, an empty tag is exported.
please correct this in order not to export empty tags, i.e. to exclude empty tags from output OpenAPI specification document.

IInt. issue: 2348

### Version

_No response_

### Logs

_No response_

### Additional information

_No response_

makeev-pavel:
# Fix: empty/unused tags in OpenAPI group export

## Problem

When exporting an API group as **Reduced source specifications**, the resulting
OpenAPI document kept root-level `tags` that **no operation in the group uses**.

## Root cause

The two export types map to two builders:

| Export type (UI) | Build type | Path |
|---|---|---|
| Combined specification | `MERGED_SPECIFICATION` | `mergeOpenapiDocuments` → `getUsedTags` (filters tags) |
| Reduced source specifications | `REDUCED_SOURCE_SPECIFICATIONS` | `DocumentGroupStrategy.transformDocumentData` (**no filtering**) |

`transformDocumentData` copied the source root `tags` verbatim (`...restOfSource`)
while rebuilding `paths` to only the group's operations — so unused tag
definitions were never dropped.

### DIFF
--- a/src/strategies/document-group.strategy.ts
+++ b/src/strategies/document-group.strategy.ts
@@ -28,6 +28,7 @@ import {
 import {
   calculateRestOperationId,
   EXPORT_API_TYPE_FORMATS,
+  filterUsedTags,
   fromBase64,
   isValidHttpMethod,
   normalizeGraphQL,
@@ -192,6 +193,9 @@ function transformDocumentData(versionDocument: VersionDocument): OpenAPIV3.Docu
     paths: {},
   }
 
+  // Tag names used by the group's operations (from normalized ops → $ref-safe).
+  const usedTagNames = new Set<string>()
+
   for (const path of Object.keys(normalizedPaths)) {
     const sourcePathItem = sourcePaths[path]
     const normalizedPathItem = normalizedPaths[path]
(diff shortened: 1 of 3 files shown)


vOrigins (approved):

b41ex (commented):
Consider adding a test for a group containing two operations with different tags (union of tag definitions kept), and one for an operation with no tags field (all root tags removed).

makeev-pavel:
> Consider adding a test for a group containing two operations with different tags (union of tag definitions kept), and one for an operation with no tags field (all root tags removed).

Added
1. 'should keep the union of tags when the group has operations with different tags'
https://github.com/Netcracker/qubership-apihub-api-processor/pull/101/changes#diff-c0b9ea43dce53abf1d8a85352f31c1d5e58d25b19c62b2414e15bdadd67c80d2R152
2. 'should drop all tags when the group operation has no tags field'
https://github.com/Netcracker/qubership-apihub-api-processor/pull/101/changes#diff-c0b9ea43dce53abf1d8a85352f31c1d5e58d25b19c62b2414e15bdadd67c80d2R172

b41ex (approved):

PR #104, opened 2026-07-21
fix: preserve colliding operations in changelog

### LINKED ISSUE #724: [Bug][BE] Builder not rendering some changes in rare cases
### Describe the bug

Builder not rendering some changes in rare cases

### To Reproduce

1.Go to portal
2.Upload spec before version [prev_ver.yml](https://github.com/user-attachments/files/30007372/prev_ver.yml)
3.Upload spec after version [current_ver.yml](https://github.com/user-attachments/files/30007371/current_ver.yml)
4.Compare
ER: all number of changes the same
AR: number of changes different
Image

### Version

_No response_

### Logs

_No response_

### Additional information

_No response_

makeev-pavel:
### Root cause
Distinct but same-shape fully-parametrized paths normalize to the **same** operation id
(e.g. `/{a}/{b}` and `/{c}/{d}` both → `*-*-get`). The changelog keyed operations by that normalized id, so one operation overwrote the other in the operations map — its changes became invisible or got misattributed.

## What changed

- **`normalizeOperationIds` (`compare.utils.ts`)** — operations are now grouped by their normalized id first:
  - **unique** normalized id → kept as the key (rename-tolerant: a renamed path parameter still matches its counterpart);
  - **colliding** normalized id → each operation is keyed by its unique **raw** `operationId`, so none is dropped.
- **New `resolveOperationPair` helper** — variadic lookup that tries candidate ids in order (normalized first, then raw), returning an empty pair when nothing matches. Replaces the previous inline `?? ... ??` fallback chain.
- **`rest.changes.ts` (`compareDocuments`)** — computes the raw operation ids (`calculateRestOperationId`) and resolves the pair through `resolveOperationPair`. This is required by the raw-id keying above; without it the normalized lookup misses on collisions and the build fails with `Can't find operation`.
- **`OperationsMap` type** widened to `Record<NormalizedOperationId | OperationId, OperationPair>`.

## Behaviour notes

- A path-parameter rename with a **unique** normalized id is matched as **one changed operation** (not remove + add).
- A rename **inside a collision group** is inherently ambiguous, so it is reported as remove + add — the deterministic, safe choice.

### DIFF
--- a/src/apitypes/rest/rest.changes.ts
+++ b/src/apitypes/rest/rest.changes.ts
@@ -18,6 +18,7 @@ import { RestOperationData } from './rest.types'
 import {
   areDeprecatedOriginsNotEmpty,
   calculateNormalizedRestOperationId,
+  calculateRestOperationId,
   isEmpty,
   isOperationRemove,
   isValidHttpMethod,
@@ -80,6 +81,7 @@ import {
   createOperationChange,
   getOperationTags,
   OperationsMap,
+  resolveOperationPair,
 } from '../../components'
 import { createRestApiCompatibilityScopeFunction } from '../../components/compare/rest.bwc.validation'
 import { calculateApiKindFromLabels, getApiKindProperty } from '../../components/document'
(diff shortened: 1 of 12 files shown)


b41ex:
Won't fix, see comment: https://github.com/Netcracker/qubership-apihub/issues/724#issuecomment-5047464441

PR #106, opened 2026-07-28
feat: sort build result lists deterministically

### LINKED ISSUE #733: [FE] Ensure build result lists are sorted
Build result includes a number of lists, at least some of them are now not ordered deteministically. It means that when some package version is rebuild during migration process, the build result could change, although the change itself is non-significant reorder of elements in the collections.
This result in suspicious build reported and additional time spent on analysis of suspicios builds.
In scope of this issue we want to ensure that all lists in build result have deterministic order.

makeev-pavel:
## The build result, field by field

Everything `createVersionPackage` writes is listed below — nothing in the package is left undecided.

### Sorted

| file in the package | field | sorted by | test |
|---|---|---|---|
| `documents.json` | `documents[]` | `fileId` | integration |
| `operations.json` | `operations[]` | `operationId` | integration |
| `comparisons.json` | `comparisons[]` | `(packageId, version, revision, previousVersionPackageId, previousVersion, previousVersionRevision)` | unit ×2 (leading field, later fields incl. numeric revision) |
| `comparisons/<comparisonFileId>` | `operations[]` | `(operationId, previousOperationId)` | integration + unit + unit (special chars in ids) |
| `comparisons/<comparisonFileId>` | `operations[].changes[]` | `calculateChangeId` | unit |
| `ddl.json` | every kind (`tables`, …) | `ddlEntityId` | integration |
| `ddl-comparisons.json` | `comparisons[]` | same comparison key as above | unit |
| `ddl-comparisons/<comparisonFileId>` | `entities[]` | `(ddlEntityId, previousDdlEntityId)` | unit |
| `ddl-comparisons/<comparisonFileId>` | `entities[].changes[]` | `calculateChangeId` | unit |
| `mcp.json` | every kind (`inits`, `tools`, `resources`, `prompts`) | `mcpEntityId` | integration (tools) + unit (prompts) |
| `version-internal-documents.json` | `documents[]` | `id` | unit (+ a filter test) |
| `comparison-internal-documents.json` | `documents[]` | `id` | unit (+ a filter test) |
| `notifications.json` | `notifications[]` | `(severity, message, fileId, operationId, previousOperationId)` | unit ×2 (first two fields, then the tail) |

### Not sorted — and why

| file in the package | field | why it is left as is |
|---|---|---|
| `operations.json` | `operations[].tags[]` | spec parse order, already deterministic |
| `operations.json` | `operations[].models` | json-crawl order; object keys, no semantics |
| `operations.json` | `operations[].deprecatedInPreviousVersions[]` | already deterministic **and** a text sort is wrong for versions (`v10 < v2 < v9`) — pinned by a guard test |
| `operations.json` | `operations[].deprecatedItems[]` | already deterministic, never a source of drift |
| `documents.json` | `documents[].operationIds[]` | `buildOperations` parse order, already deterministic |
| `info.json` | `config.refs[]` | comes verbatim from the caller's `BuildConfig`, so already deterministic — and the order is consumed positionally when resolving dashboard/group documents, so sorting it would change which document wins a slug collision |
| `operations/<operationId>` | operation payload | a single object; key order comes from the spec |
| `documents/<filename>` | dumped document | a single document (JSON / SDL / SQL) |
| `version-internal-documents/<id>.json` | serialized document | a single serialized object |
| `comparison-internal-documents/<id>.json` | serialized comparison document | a single serialized object |
| `mcp/<mcpEntityId>`, `ddl/<ddlEntityId>` | entity payload | a single object / raw SQL |
| search text files | operation search text | built by a deterministic `Set` crawl |
| `info.json` | everything except `refs` | scalar config fields |

**Principle behind the split:** sort only lists whose order came from `Map` iteration or async completion. Sorting an already deterministic field is over-reach — it would produce a one-time false "changed" diff on the first rebuild after merge, which is exactly the false positive this issue exists to kill.

## What changed in the tests

**Packaging in tests is now the same packaging as in production.** Previously the test registries had their own serializer (~265 lines of `save*` / `get*Content`) that assembled the version directory itself — a second implementation that could, and did, drift away from `createVersionPackage`. Now `LocalRegistry.publishPackage` and `ApihubRegistry.publishPackage` both delegate to `createVersionPackage` through a filesystem-backed `ZipTool`, and that parallel serializer is deleted. Tests read exactly the file layout the product writes, so ordering assertions are made against the real artifact rather than a look-alike.

That switch immediately exposed three defects that the duplicate serializer had been hiding:

- **operation data was silently lost on reload** — the test registry read `operations/<id>.json` while production writes `operations/<id>` with no extension, so `operation.data` came back `undefined` (proven on a real filesystem with a negative control);
- **index and data files could disagree** — the comparison-internal index filtered on `comparisonFileId` while the writer filtered on `serializedComparisonDocument`, so an index row could point at a file that was never written; both now go through one resolver;
- **fire-and-forget writes** — `createVersionPackage` calls `zip.file(...)` without `await` (JSZip's `file()` is synchronous), so the async sink tracks writes in a pending queue drained by the always-awaited `buildResult()`. `memfs` hid this race; `FS_MODE=disk` exposed it.

New test file `test/build-result-ordering.test.ts`, 18 tests in two blocks:

- **4 integration cases** load the real zip produced by `createVersionPackage` and assert the exact serialized order. Fixtures declare files, paths and tables in reverse-alphabetical order, so a dropped sort fails;
- **14 unit cases** feed each exported builder reverse-order input and assert the sorted output — they cover the lists an integration build does not reach directly.

Every test carries a `// before: … → after: …` line stating the input order and the resulting order.

## Design decisions worth reviewing

- **Comparison is by UTF-16 code unit (`<` / `>`), never `localeCompare`** — locale and ICU build differ between Node and the browser, which is exactly the cross-runtime nondeterminism this issue is about.
- **`sortByKey` is decorate-sort-undecorate**, so each key is computed once per item instead of on every comparison; the priciest key (`calculateChangeId`) sorts 1998 real changes in 6.3 ms.
- **Numbers inside composite keys are zero-padded** so a text comparison keeps numeric order (revision 2 before 10); values outside non-negative integers below 1e10 throw instead of sorting wrongly.
- **`changes[]` is sorted** because for multi-document comparisons an operation's changes are concatenated across document pairs and then deduplicated — that is aggregation order, not source order. Display is unaffected: the portal re-sorts changes itself.
- **Notifications are sorted** because they are emitted from concurrently processed files (`Promise.all`), so their order is completion order. The key spans every field of `NotificationMessage`, so equal keys mean byte-identical rows.

## Validation

- `tsc --noEmit` clean; eslint 0 errors.
- Full suite green in both modes: `jest` and `npm run test:disk` — 68 suites / 1249 passed.
- Negative controls run for the non-obvious fixes (operation-data reader, index/data filter, sort wiring).

### Performance

A/B of the full test suite, develop vs this branch — 8 runs with interleaved rounds, median of 3 measured runs per side, one warmup discarded:

| metric | develop | this branch |
|---|---|---|
| suite wall clock (median) | 362.7s | 347.5s |
| sum of per-file durations | 1678.9s | 1662.7s |
| same, common files only | 1678.9s | 1655.0s |

**No measurable difference.** Medians favour this branch by 4%, minimums favour develop by 1% — the two metrics disagree in direction, so this is run-to-run noise (three runs on unchanged code spanned 679 / 363 / 343 seconds). The only real figure: the 18 new tests cost about 7.7s of CPU work, a few seconds of wall clock across three workers.


makeev-pavel:
Tests ✅
https://github.com/Netcracker/qubership-apihub/actions/runs/30429056816/job/90501828696

b41ex (changes requested):
Add corresponding agent instructions to make sure build result lists stay sorted deterministically in the future. Add `api-processor-authoring` agent package and put it there.

test/publish-package.test.ts
```diff
@@ -35,12 +35,14 @@ describe.skip('Publish Package Test', () => {
   })
 
   test('Update package operations dataHash', async () => {
```
b41ex: Let's just remove the test, no need to keep it. We got rid of calculating operations hash some time ago.

b41ex (changes requested):

test/publish-package.test.ts
makeev-pavel: done

b41ex (approved):

PR #107, opened 2026-07-28
docs: plan for split APIHUB validation results by specification type

b41ex:


docs/plans/tolerant-publication-critical-errors.md
```diff
@@ -0,0 +1,759 @@
...
+| `documents.json` — each document | new optional `hasErrors` (default `false`) |
+| `info.json` — the version | new optional `hasErrors` (default `false`) |
+| `notifications.json` | unchanged location; entries follow the new `NotificationMessage` shape |
+
+**Backend — REST API**
+
+| Endpoint | Change |
+|----------|--------|
+| `GET /api/v3/packages/{packageId}/versions` | `hasErrors` per version |
+| `GET /api/v3/packages/{packageId}/versions/{version}` | `hasErrors` for the version; `hasErrors` per `operationTypes` entry and per `contractsSummary.ddl` / `.mcp` |
```
karpov-aleksandr, line 84: `contractsSummary.mcp` is an endpoint-keyed map, so adding `hasErrors` flag is not possible without contract changes. We can change the format of `contractSummary.mcp` and make it an object, and move the current map under the `contractSummary.mcp.endpointsSummary` and then we can add the `hasErrors` flag, but this is breaking change. Or we can make an additive change and add to `contractSummary` something like that `contractTypesWithErrors: ["ddl","mcp"]`

docs/plans/tolerant-publication-critical-errors.md
```diff
@@ -0,0 +1,759 @@
...
+| `GET /api/v2/packages/{packageId}/versions/{version}/documents` | `hasErrors` per document |
+| `GET /api/v3/packages/{packageId}/versions/{version}/documents/{slug}` | `hasErrors` for the document |
+| **new** `GET /api/v2/packages/{packageId}/versions/{version}/notifications` | version's build notifications, filterable by `documentId`, `severity`, `category` |
```
karpov-aleksandr, line 87: If we want to show notifications for comparisons too, we also need to add the `previousVersionPackageId` and `previousVersion` parameters. Also a scope type parameter with possible values: ["version", "comparison"] would be useful to have a clear understanding of how to search for notifications.

docs/plans/tolerant-publication-critical-errors.md
```diff
@@ -0,0 +1,759 @@
...
+## Decisions still to confirm
+
+1. Final `MessageCategory` value set (proposed above) — confirm during the severity review.
+2. Per-message severity confirmations (the **review** rows in §2) — human decision, DB-analysis driven.
+3. Notification id in the api-processor shape — recommended **not** to add (see §1).
+4. Changelog-phase notifications (`rest.changes.ts:269/278`, `package.ts:64`, `deprecated.ts:141/149`): keep as
+   `Error` and let them count toward the current version's `hasErrors`, downgrade to `Warning`, or exclude the
+   `Changelog`/`Packaging` categories from the `hasErrors` computation? Three options weighed in §5;
+   recommendation is to downgrade the risky-validation pair. Needs the review.
+5. Whether `hasErrors` should be written into the changelog build result at all. A `changelog` build produces no
```
karpov-aleksandr, line 751: Currently, the backend only reads builder notifications for `type=build`; they are not processed for `type=changelog`. This is a gap on BE side. Actually, I don't see a problem with marking the comparison with `hasErrors` in `comparisons.json`or `ddl-comparisons.json` and storing this information as a version comparison metadata. And we can also return `hasErrors` in the response for comparison-related endpoints.

docs/plans/tolerant-publication-critical-errors.md
```diff
@@ -0,0 +1,759 @@
...
+
+- `operationTypes.*.hasErrors` and `contractsSummary.*.hasErrors` are computed by the backend from stored
+  per-document `hasErrors`. Two cases produce a version-level mark with no API type highlighted, and the UI must
+  tolerate both (see §2b): a document that failed before its type could be determined maps to no
+  apiType/contractType; and a version-level `Error` (missing previous version, reference resolution) flags no
+  document at all. In both cases `hasErrors` on the version is the only signal, and the notifications endpoint
+  is where the explanation lives.
+- Flags are per revision (both tables are revision-keyed); the mark disappears when a fixed revision is
+  published.
+- Dashboard viewing: `GET .../documents` resolves referenced packages' documents, so per-document `hasErrors`
```
karpov-aleksandr, line 681: The design above states that the dashboard cannot have references to such packages.

docs/plans/tolerant-publication-critical-errors.md
```diff
@@ -0,0 +1,759 @@
...
+| `NotificationMessage` | `fileId` → `documentId` (a single document slug); new required `category`; `operationId` / `previousOperationId` removed. A problem concerning several documents yields one message per document. |
+| `documents.json` — each document | new optional `hasErrors` (default `false`) |
+| `info.json` — the version | new optional `hasErrors` (default `false`) |
+| `notifications.json` | unchanged location; entries follow the new `NotificationMessage` shape |
+
+**Backend — REST API**
+
+| Endpoint | Change |
+|----------|--------|
+| `GET /api/v3/packages/{packageId}/versions` | `hasErrors` per version |
```
karpov-aleksandr, line 83: We also should add additional validation to `PATCH /api/v2/packages/{packageId}/versions/{version}` to not allow to change a version status that marked with `hasErrors` flag.

docs/plans/tolerant-publication-critical-errors.md
b41ex: > `contractsSummary.mcp` is an endpoint-keyed map, so adding `hasErrors` flag is not possible without contract changes

I do not not follow how the second statement follows from the first. Adding property to the response is backward compatible and is reported as non breaking by APIHUB.

b41ex: > `contractsSummary.mcp` is an endpoint-keyed map, so adding `hasErrors` flag is not possible without contract changes

I do not not get how the second statement follows from the first. Adding property to the response is backward compatible and is reported as non breaking by APIHUB.

docs/plans/tolerant-publication-critical-errors.md
b41ex: Yes, it is prohibited by the design to add such packages to the dashboard. This is a precaution for the following case- suppose we created a dashboard and added a (valid) package version to it. After it we deployed a new version of APIHUB with stricter validations and the package version (already included in the dashboard) now has errors. It is not a typical scenario, but could happen. The least we can do in this case is indicate that there's a problem with the package version included in the dashboard.

docs/plans/tolerant-publication-critical-errors.md
b41ex: Agree

docs/plans/tolerant-publication-critical-errors.md
karpov-aleksandr: Perhaps I didn't understand your initial idea quite correctly. The `mcp` field contains a map in which the `mcp endpoint` is the key and `mcpEndpointSummary` is the value. If your idea is to add `hasErrors` for each `mcp endpoint`, then yes, we can add `hasErrors` to `mcpEndpointSummary` and it won't be a breaking change. Please clarify.

docs/plans/tolerant-publication-critical-errors.md
karpov-aleksandr: Agree, we should handle such a case

docs/plans/tolerant-publication-critical-errors.md
```diff
@@ -0,0 +1,759 @@
...
+produces **one notification per affected document**, all carrying the same `message`, `category` and
+`severity`, each with its own `documentId`.
+
+This keeps every consumer simple: `hasErrors` per document is a plain lookup, the endpoint's `documentId`
+filter is a plain equality match, and the backend stores one row per notification with no splitting or array
+containment. The cost is repeated message text, which is a fair trade — each copy is individually meaningful,
+because a user looking at one document wants that document's problems regardless of what else the problem
+touches.
+
+**fileId→slug resolution.** Some notifications are raised where only `fileId` is known (parse-time,
```
makeev-pavel: # Attribute notifications with the slug from the start

**Proposal:** put the document **slug** into `documentId` at the site that raises the notification, instead of
raising with `fileId` and converting in one pass at package-creation time.

## Why

`fileId` and slug are different strings — `createFileSlugs` strips the base path and extension, slugifies, and
resolves collisions:

| `fileId` | slug |
|---|---|
| `specs/petstore.yaml` | `petstore` |
| `specs/v2/petstore.json` | `petstore-1` (when `petstore` is taken) |

Both are plain strings, so a mix-up is silent. A `fileId` that reaches `documentId` becomes a document id no
document has: `?documentId=petstore` returns nothing and the document's `hasErrors` is never set. Converting
late also means the field holds one kind of value for most of the build and another at the end, which anything
reading `BuildResult` in memory sees.

## It is cheap

Slugs are assigned to every file before parsing starts, so they exist before the first document-level
notification. Of the ten sites that carry a document identifier, **nine already have the slug in scope** — one
argument each. Three of them destructure it on the previous line and pass `fileId` anyway:

```ts
const { fileId, slug = '', publish = true, apiKind, ...fileMetadata } = file
const { data, dependencies } = await getBundledFileDataWithDependencies(
  fileId, ctx.parsedFileResolver, createBundlingErrorHandler(ctx, fileId),   // slug is right there
)
```

## One site stays a problem, either way

`builder.ts:755` raises `Invalid <type> file` from `parseFile(fileId, source)`, where the slug can be genuinely
absent: the bundler also parses files pulled in through `$ref`, and such a file may not be in `config.files` at
all. It never becomes a document, so it has no slug. Leaving `documentId` unset is the right outcome — the
field is optional and such a notification is a version-level error, which both designs describe.

Late conversion does not solve this site either: its `fileId → slug` map has no entry for that file, so the
`fileId` survives the pass and ships as if it were a slug. Attributing at the source makes the absence
explicit instead.

docs/plans/tolerant-publication-critical-errors.md
```diff
@@ -0,0 +1,759 @@
...
+### 3. Catch points (make Groups A and B tolerant)
+
+- `components/files.ts` `buildFile`: wrap `parsedFileResolver` and `buildDocument` in `try/catch`; on throw
+  push an `Error` notification (category `DocumentParsing`/`DocumentBuild`) and return `buildErrorDocument(file)`
+  so the file stays visible as `type: unknown` with a slug.
+- `BuildStrategy.execute` per-document loop: wrap `processOperationDocument`, `validateDdlDocument` +
+  `processDdlDocument`, `processMcpDocument` per document → `Error` notification, continue. The failed document
+  stays published but exposes no operations/entities.
+- duplicate handlers (B1–B5): emit one `Error` notification **per involved document**, same text, and keep the
+  first-seen entity/operation (do not overwrite); do not throw.
```
makeev-pavel: First-seen entity/operation wins: 
- Rebuilding the same sources can change the published content. If the config is assembled in a different order — different upload order, a changed discovery order on the agent side — the other copy wins. Nothing in the version changed, but the operation did.

- The changelog will report that as a real change. The operation's content and its owning document both differ from the previous version, so the comparison shows a modification no one made.

docs/plans/tolerant-publication-critical-errors.md
```diff
@@ -0,0 +1,759 @@
...
+**Consequence to accept:** the four version-level rows mean a version can have `hasErrors: true` while every
+document has `hasErrors: false`. The UI must handle that — the version-level mark cannot assume a document to
+point at, and the notifications endpoint (unfiltered) is the only place those messages surface. This is correct
+behavior, not a gap: a missing previous version really is a version-level problem.
+
+### 3. Catch points (make Groups A and B tolerant)
+
+- `components/files.ts` `buildFile`: wrap `parsedFileResolver` and `buildDocument` in `try/catch`; on throw
+  push an `Error` notification (category `DocumentParsing`/`DocumentBuild`) and return `buildErrorDocument(file)`
+  so the file stays visible as `type: unknown` with a slug.
```
makeev-pavel: `buildErrorDocument `save document as `data: ''` (empty string). We need to put file in data?

docs/plans/tolerant-publication-critical-errors.md
b41ex: Yes. See https://github.com/Netcracker/qubership-apihub-backend/pull/497/files with specific proposal for API changes.

docs/plans/tolerant-publication-critical-errors.md
b41ex: Both parameters are in. We addressed the scope concern differently, though — with two endpoints instead of a `scope` parameter:

 - `GET /api/v2/packages/{packageId}/versions/{version}/notifications` — build-phase messages, filterable by `documentId`, `severity`, `category`.
 - `GET /api/v2/packages/{packageId}/versions/{version}/changes/notifications` — comparison messages, the same filters **plus** `previousVersion` and `previousVersionPackageId`.

 The reasoning is that the two identify different subjects: a build notification is addressed by *(packageId, version)*, a comparison notification by *(packageId, version, previousVersionPackageId, previousVersion)* — the same key `/changes/summary` already uses. With one endpoint, `previousVersion` becomes conditionally required depending on `scope`, and callers have to disambiguate results by inspecting them. Splitting makes the scope structural instead of a parameter. Happy to revisit if you'd rather have the single endpoint.

Plan: Backend §3; spec: both paths in `APIHUB_API.yaml`.

docs/plans/tolerant-publication-critical-errors.md
b41ex: Agreed on all three, and all three are in the plan. Our own data analysis independently confirmed the gap you flagged: `PublishChanges` never calls `ReadBuilderNotificationsToEntities`, which is why `previous-version-missing` shows zero occurrences in production despite being reachable.

 - **`hasErrors` on comparisons** — added to `BuildResultComparisonData` in `APIHUB_API_internal.yaml`, which both `comparisons.json` and `ddl-comparisons.json` entries extend, so it is one addition covering both.
 - **Stored as comparison metadata** — persisted on the `version_comparison` row (Backend §2), so the flag is served without aggregating notifications.
 - **Returned by comparison endpoints** — `hasErrors` on `/changes/summary` (both the package-comparison shape and per `refs[]` entry for dashboards) and `comparisonHasErrors` on version content with `includeSummary=true`.

 For the notifications themselves we added a second build-result file, `comparison-notifications.json`, and a `comparison_notifications` table keyed by comparison id, written by **both** publish paths — so the changelog-build gap closes rather than persisting.

Plan: [`hasErrors` flags in the build result](docs/plans/tolerant-publication-critical-errors.md), [Every comparison notification belongs to one version pair](docs/plans/tolerant-publication-critical-errors.md), Backend §1–§3.

b41ex: Agreed on all three, and all three are in the plan. Our own data analysis independently confirmed the gap you flagged: `PublishChanges` never calls `ReadBuilderNotificationsToEntities`, which is why `previous-version-missing` shows zero occurrences in production despite being reachable.

 - **`hasErrors` on comparisons** — added to `BuildResultComparisonData` in `APIHUB_API_internal.yaml`, which both `comparisons.json` and `ddl-comparisons.json` entries extend, so it is one addition covering both.
 - **Stored as comparison metadata** — persisted on the `version_comparison` row (Backend §2), so the flag is served without aggregating notifications.
 - **Returned by comparison endpoints** — `hasErrors` on `/changes/summary` (both the package-comparison shape and per `refs[]` entry for dashboards) and `comparisonHasErrors` on version content with `includeSummary=true`.

 For the notifications themselves we added a second build-result file, `comparison-notifications.json`, and a `comparison_notifications` table keyed by comparison id, written by **both** publish paths — so the changelog-build gap closes rather than persisting.

Plan: [`hasErrors` flags in the build result](docs/plans/tolerant-publication-critical-errors.md), [Every comparison notification belongs to one version pair](docs/plans/tolerant-publication-critical-errors.md), Backend §1–§3.

docs/plans/tolerant-publication-critical-errors.md
b41ex: Both halves are covered. Publishing a dashboard that references an unsound version is refused going forward, and the pre-existing case you described — a version that became errored under stricter validation *after* it was added — is handled by indication rather than refusal, since the reference already exists.

 The documents-list endpoint resolves referenced packages' documents, so per-document `hasErrors` propagates into the dashboard document list automatically; the UI requirement is recorded under UI follow-up → Other screens.

 One refinement worth noting: the refusal gates on an **unsound** version — errors of its own **or** an unreliable changelog — the same predicate used by the other three refusals, so they cannot drift apart. Spec: `ReferencedVersionHasErrors` example on `POST /api/v2/packages/{packageId}/publish`.

docs/plans/tolerant-publication-critical-errors.md
b41ex: Added, scoped to promotion into `release` only. Other transitions — notably `draft` → `archived` — stay allowed so an unsound version can still be cleaned up.

 It belongs inside the existing `if newStatus == string(view.Release)` block in `VersionService.go`, alongside the release-version-pattern and previous-version-status checks already there, and it reads the stored flags rather than recomputing anything. Like the other guards it gates on the unsound predicate, so a version whose *changelog* has errors is also blocked from promotion.

 Spec: `VersionStatusChangeWithErrors` 400 example on the PATCH endpoint. Plan: Backend §2 and [Public API changes at a glance](docs/plans/tolerant-publication-critical-errors.md) → Backend — new refusals.

docs/plans/tolerant-publication-critical-errors.md
b41ex: Adopted — no late conversion pass. Every site that knows its document passes `document.slug` directly ([Message shape](docs/plans/tolerant-publication-critical-errors.md)).

 Two things came out of working through it:

 **The `$ref` site is attributable after all.** Rather than leaving `invalid-text-file` unattributed, it is attributed to the **root document** that pulled the file in — its bundle is what ends up broken. Emission moves out of `parseFile` into the document build, which reads `SourceFile.errors` for its own file *and each of its `dependencies`*, so a file referenced by two documents produces one notification per root instead of one attributed to whichever parsed first. The message names the offending `fileId` in its text, since it is now attached to a different document. ([Errors in `$ref`-ed files](docs/plans/tolerant-publication-critical-errors.md))

 **`documentId` already meant two different things.** `ApiOperation.documentId` is a slug, but `McpEntity.documentId` and `DdlEntity.documentId` hold a `fileId` — while both the internal and public specs advertise a slug for MCP, and the public spec advertises one for DDL too. So the mix-up you warned about had already shipped. Unifying them on the slug is now part of this change. ([`documentId` Unification (MCP and DDL Entities)](docs/plans/tolerant-publication-critical-errors.md))

 A whole-result invariant guards it: every `documentId` emitted anywhere — notifications, `operations.json`, `mcp.json`, `ddl.json` — must match a `slug` in `documents.json` (Tests → T5).

docs/plans/tolerant-publication-critical-errors.md
b41ex: Agreed, and fixed. Two corrections to the premise while we were in there: the current behaviour is **last**-wins, not first-wins — `setReportingDuplicate` calls `map.set` unconditionally after invoking the handler — and processing order is `config.files` order, which is deterministic for a given config but carries no meaning.

 The decision is to keep the entity whose `documentId` sorts **first lexicographically**, so the same file set produces the same published index however the config lists them. The notification is still emitted for *both* documents; the tie-break decides only which entity is indexed, never what is reported.

 Your second point needed a separate fix. `processOperationDocument` sets `document.operationIds` to every operation it built, winner or not, so a losing document still lists the shared id — and rebuilding that loser via `update()` evicts the winner's entry before re-adding its own, flipping the winner with no content change. The delete pass is scoped to entries the document actually owns.

 This does change today's outcome for existing REST/GraphQL duplicates, which we judged acceptable because a duplicate now flags both documents and blocks `release` — a loud state rather than a silent one. ([Duplicate resolution](docs/plans/tolerant-publication-critical-errors.md), Tests → T8)

docs/plans/tolerant-publication-critical-errors.md
b41ex: Yes — and it turned out to be mandatory, not just desirable. `buildErrorDocument` produces `type: unknown`, which resolves to `unknownApiBuilder`, whose dumper **throws** when the document has no `source`. Error documents have `publish: true`, so they are not skipped. As the plan stood, every tolerated document would have died in `createVersionPackage` — the build would catch the parse error, record a good notification, assemble a draft with all the healthy documents, and then fail at packaging.

 What we save is `source` (the raw bytes), not `data`. `data` is the parsed model and feeds operations, search and `/files/{slug}/doc`; for an unparseable file there is no meaningful parsed model, so `data: ''` stays correct. `source` is what lands in `documents/<filename>` and is served by `/files/{slug}/raw` — the broken file itself, downloadable, which is the actual troubleshooting artifact.

 Also: `dumpUnknownDocument` now returns an empty `Blob` instead of throwing, because one case remains where there genuinely is no source (the file could not be fetched at all). Setting `publish: false` was rejected — it would hide the failed document entirely. ([Error documents must carry their source](docs/plans/tolerant-publication-critical-errors.md), Tests → T2)

docs/plans/tolerant-publication-critical-errors.md
```diff
@@ -0,0 +1,1790 @@
...
+- **Dashboards:** an unsound package version **cannot be added** to a dashboard. If a dashboard somehow already
+  references one, the UI must indicate it when viewing the dashboard's packages.
+
+UI implementation is **out of scope**; UI source was analyzed only to identify the backend APIs that change.
+
+## High-level design
+
+1. **Publication becomes tolerant per document.** A failure while processing one document no longer aborts the
+   build. It is recorded against that document; the document is still published — carrying its original
+   content, so the user can inspect it, but exposing no operations or entities — and every other document
```
makeev-pavel: `exposing no operations or entities`
But does the duplication error still publish the operations?

PR #108, opened 2026-07-28
fix: build a changelog for versions published from a GraphQL introspection

### LINKED ISSUE #742: [FE] Can't publish GQL spec from test-service with previous version set, but without previous versions in works
Precodnitions:

1. Deploy APIHUB
2. Deploy APIHUB Agent
3. Deploy https://github.com/Netcracker/qubership-apihub-test-service/tree/main/helm-template
4. Create a workspace A, create a package in it, set service name on package to `qubership-apihub-test-service`

Steps to reproduce:

1. Go to agent tab, choose workspace A, agent and namespace with qubership-apihub-test-service
2. Run discovery, make snapshot with version 1111.1 and no previous version.
3. Go till the end of the wizard and promote the version as Release
4. Refresh the page, restart discovery
5. Create snapshot with version 1111.2, choose 1111.1 as previous version 
6. Hit Create Snaphot

AR: weird shapshot error
ER: no errors

Image

makeev-pavel:
## Why

Publish a GraphQL introspection response as a release, then publish a second release with the first one as
its previous version, and the build crashes:

```text
TypeError: Cannot destructure property '__schema' of 'undefined' as it is undefined.
    at buildFromIntrospection (@netcracker/qubership-apihub-graphapi)
    at buildGraphQLDocument (src/apitypes/graphql/graphql.document.ts:34)
    at compareDocuments (src/apitypes/graphql/graphql.changes.ts)
```

The first publication succeeds. The failure happens on the second one, while comparing the two versions —
which is the agent-wizard snapshot flow reported in the issue.

A `graphapi`-sourced document hit the same mismatch without crashing: it fell into the `else` branch, where a
`GraphQLSchema` was handed to code expecting a `GraphApiSchema`, and the comparison ran over the wrong
structure. Only SDL sources were accidentally correct.

## Root cause

A GraphQL document is **always stored as printed SDL** — `dumpGraphQLDocument` runs `printGraphApi` and
writes `<slug>.graphql` — whatever the source was. But `documents.json` kept the metadata of the **source**:
for an introspection response that is `type: introspection`, `format: json`.

`compareDocuments` rebuilt the document through `buildGraphQLDocument`, spreading the stored document to fill
in the remaining fields. The spread carried the stale `type`, so the introspection branch looked for
`__schema` inside a `GraphQLSchema`. `'__schema' in schema` is `false` and `schema.data` is `undefined`, hence
the crash inside graphapi.

## What changed

**The document metadata now describes the artifact, not the source.** `buildGraphQLDocument` records
`type: graphql-schema`, `format: graphql` and derives the filename extension from the same constant. SDL is
what actually gets stored, so this is simply the truth about the file on disk — and it keeps working if the
real source document is persisted instead one day.

**Reading a stored document goes by those persisted fields.** `graphql.changes.ts` parses the raw file
directly instead of rebuilding a document to get at its model:

```ts
let prevDocData = prevFile && prevDoc && parseGraphQLDocument(await prevFile.text(), prevDoc.type, prevDoc.format)
let currDocData = currFile && currDoc && parseGraphQLDocument(await currFile.text(), currDoc.type, currDoc.format)
```

`parseGraphQLDocument` pairs `parseDocumentSource(source, format)` — SDL, JSON or YAML — with
`toGraphApiSchema(data, type)`, which is the branching `buildGraphQLDocument` already had, now extracted and
shared by the write and read paths. Both `default` branches **throw** rather than cast: the silent cast is
what produced a meaningless changelog instead of an error.

This drops the fake `TextFile` the old code assembled (whose `format` field also lied), a duplicated
`buildSchema` call and four imports.

### DIFF
--- a/src/apitypes/graphql/graphql.changes.ts
+++ b/src/apitypes/graphql/graphql.changes.ts
@@ -15,7 +15,7 @@
  */
 
 import { calculateGraphqlOperationId, isEmpty, takeIf } from '../../utils'
-import { parseGraphQLSource } from '../../utils/graphql-transformer'
+import { parseGraphQLDocument } from './graphql.document'
 import {
   aggregateDiffsWithRollup,
   apiDiff,
@@ -69,9 +69,9 @@ export const compareDocuments: DocumentsCompare = async (
   const comparisonInternalDocumentId = createComparisonInternalDocumentId(previousVersion, previousPackageId, prevDoc?.slug, currentVersion, currentPackageId, currDoc?.slug)
   const prevFile = prevDoc && await rawDocumentResolver(previousVersion, previousPackageId, prevDoc.slug)
   const currFile = currDoc && await rawDocumentResolver(currentVersion, currentPackageId, currDoc.slug)
-  // The raw document is SDL whatever the version was published from, so parse it as a schema.
-  let prevDocData = prevFile && parseGraphQLSource(await prevFile.text())
-  let currDocData = currFile && parseGraphQLSource(await currFile.text())
+
+  let prevDocData = prevFile && prevDoc && parseGraphQLDocument(await prevFile.text(), prevDoc.type, prevDoc.format)
+  let currDocData = currFile && currDoc && parseGraphQLDocument(await currFile.text(), currDoc.type, currDoc.format)
 
   if (!prevDocData && currDocData) {
     prevDocData = getCopyWithEmptyOperations(currDocData)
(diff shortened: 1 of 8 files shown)


makeev-pavel:
Tests ✅
https://github.com/Netcracker/qubership-apihub/actions/runs/30364966755/job/90293771269

b41ex (changes requested):
We should store the actual `type`/`format` for an artifact and use it during parsing.

Yes, today it is always SDL, but may be in the future we will start persisting the real source documents, not already processed.

b41ex (approved):

PR #112, opened 2026-08-07
feat: resolve api kind from xApiKind in the build config

### LINKED ISSUE #750: [FE] Support `xApiKind` specified for the file in build config
There are cases when backend composes a build config with `xApiKind` property specified for the file. 

Need to support yet another way to specify API Kind in `api-processor`.

Should be supported for OpenAPI, AsyncAPI, GraphQL specifications.

In terms of priority when calculating effective API Kind, `xApiKind` property specified for the file in build config should be placed between file label `apihub/x-api-kind` and version label `apihub/x-api-kind`.

makeev-pavel:
Support `xApiKind` in the build config

## Why

A build config can carry the api kind of a document in the `xApiKind` field of its file entry.
api-processor had no support for it: the field reached the processor through the index signature of
`BuildConfigFile` and was silently dropped, so a document marked as internal upstream was published
as `BWC`.

## What

**`xApiKind` as a third source of the document api kind.** `calculateFileApiKind` resolves one
build-config file, from the strongest source to the weakest:

| # | Source | Scope |
| - | ------ | ----- |
| 1 | file label `apihub/x-api-kind` | one document |
| 2 | `xApiKind` of the file | one document |
| 3 | version label `apihub/x-api-kind` | every document of the version |
| 4 | `BWC` | default |

A per-file source beats a per-version one, and an explicit label beats the transport field. Blank or
whitespace-only means nothing was stated and lets the version label through. Any other value states
something, and an unrecognized one states `BWC` rather than falling through, so a document marked
explicitly does not inherit the kind of the whole version. The value is trimmed, because the label
branch already trims what it extracts.

Resolution happens in `buildDocument`, the single funnel for every builder, so REST, GraphQL and DDL
pick the new source up unchanged. `consumeXApiKind` reads the field and removes it there, once, so it
never reaches the metadata of a published document. A document is built on a copy of its entry of the
build config: the entry outlives the build, and an incremental rebuild hands the very same one over
again.

### DIFF
--- a/src/apitypes/async/async.changes.ts
+++ b/src/apitypes/async/async.changes.ts
@@ -46,6 +46,7 @@ import {
   OperationsMap,
 } from '../../components'
 import { createAsyncApiCompatibilityScopeFunction } from '../../components/compare/async.bwc.validation'
+import { calculateApiKindFromLabels } from '../../components/document'
 import { v3 as AsyncAPIV3 } from '@asyncapi/parser/esm/spec-types'
 import {
   extractAggregatedDiffs,
@@ -70,6 +71,7 @@ export const compareDocuments: DocumentsCompare = async (
     currentPackageId,
     currentGroup,
     previousGroup,
+    previousVersionLabels,
   } = ctx
 
   const comparisonInternalDocumentId = createComparisonInternalDocumentId(previousVersion, previousPackageId, prevDoc?.slug, currentVersion, currentPackageId, currDoc?.slug)
@@ -86,6 +88,11 @@ export const compareDocuments: DocumentsCompare = async (
     currDocData = createCopyWithEmptyOperations(prevDocData)
   }
 
+  const currDocumentApiKind = currDoc?.apiKind
+  // ApiKind is not guaranteed for the previous document: it is downloaded data and the field is optional.
+  // In this case, we calculate ApiKind by labels (file then version) in order of priority.
+  const prevDocumentApiKind = prevDoc?.apiKind || calculateApiKindFromLabels(prevDoc?.labels, previousVersionLabels)
+
   const { merged, diffs } = apiDiff(
     prevDocData,
     currDocData,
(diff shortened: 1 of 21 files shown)


src/types/external/config.ts
```diff
@@ -209,6 +209,10 @@ export interface BuildConfigFile {
   fileId: FileId
   slug?: string // for rebuild
   apiKind?: ApihubApiCompatibilityKind
+  /**
+   * Api kind the Agent discovered for this document
```
b41ex: The comment is a little bit misleading. Agent is just a known user of this field, but not necessarily exclusive.
Also I think this is the place to document how `apiKind` (just above) and `xApiKind` are related. It is not obviuos.

src/components/document.ts
```diff
@@ -66,7 +66,7 @@ export const buildDocument = async (parsedFile: SourceFile, file: BuildConfigFil
   const apiBuilder = ctx.apiBuilders.find(({ types }) => types.includes(parsedFile.type)) || unknownApiBuilder
 
   try {
-    file.apiKind = calculateApiKindFromLabels(file.labels, ctx.versionLabels)
+    file.apiKind = calculateFileApiKind(file, ctx.versionLabels)
```
b41ex: Can we remove xApiKind from `file` right here, once we used it, instead of doing throwaway destructuring repeated in each specific document type?

src/apitypes/async/async.changes.ts
```diff
@@ -86,6 +88,11 @@ export const compareDocuments: DocumentsCompare = async (
     currDocData = createCopyWithEmptyOperations(prevDocData)
   }
 
+  const currDocumentApiKind = currDoc?.apiKind
+  // ApiKind is not guaranteed for the previous document: it is downloaded data and the field is optional.
+  // In this case, we calculate ApiKind by labels (file then version) in order of priority.
+  const prevDocumentApiKind = prevDoc?.apiKind || calculateApiKindFromLabels(prevDoc?.labels, previousVersionLabels)
```
b41ex: Shouldn't this fallback take into account xApiKind from config? It seems we are losing it if we reconstruct it only from labels, but not taking into account xApiKind .

The same question for other API types where we have similar code.

b41ex (changes requested):

src/apitypes/async/async.changes.ts
makeev-pavel: I checked it out. All information is saved in apiKind and there is no need to go into the labels. Deleted it

src/components/document.ts
makeev-pavel: fixed

src/types/external/config.ts
makeev-pavel: fixed

b41ex (approved):

